PR #3, opened 2024-06-28
Dawntrail - Fix Fire 3 Hard Cast

TMTurtle:
* Broke it in a84b246

### DIFF
--- a/src/Game/StatsModifier.ts
+++ b/src/Game/StatsModifier.ts
@@ -65,7 +65,7 @@ export class StatsModifier
 			uiMod.manaCostFire = 0;
 			uiMod.manaCostIce = 0;
 			if (ui.availableAmount() === 3)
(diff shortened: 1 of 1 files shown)


miyehn:
Gotcha ty 👌 actually I'll leave this PR open for now in case there are more changes. I'll come back after I'm done with the local storage related changes. Probably tomorrow evening (PST)

miyehn:
I found a couple more things- if you're busy I can get to them later:
- currently calling Fire and Blizzard from unaspected state doesn't grant Thunderhead
- all ice spells should `gainUmbralMana`. I haven't checked all spells but currently High Blizzard 2 doesn't
- addle duration -> 15s
- `switchToAForUI` should clear Astral Soul when switching away from Astral Fire state
- after Retrace is called in LL, might want to switch skill icon back to LL..?

PR #5, opened 2024-07-18
New Feature - Party Buff Alignment

TMTurtle:
* Add Buffs via Timeline Marker system
** Potency calculations on spell hover and damage table respect cumulative party buffs
** For Crit/DH-based buffs, character attributes must be entered in configuration

* Needs translation:
** Buff Marker Entry
** Crit/DH entry
** DamageStatistics (party)


miyehn (commented):
(haven't done looking at all changes ofc but will comment as I go-)

Buff icons: I noticed that some of them don't have the highlight/drop shadow, including LL and pot. Currently they are displayed slightly differently on the timeline and the misaligned pixels annoy me a little xD

image

We should probably keep them consistent eventually.

(I might also squish them closer together at some point.. this kind of minor style adjustments can all wait until after this pr though)

src/Game/GameState.ts
```diff
@@ -760,6 +761,29 @@ export class GameState {
 		skill.use(this, node);
 	}
 
+	getPartyBuffs(time?: number) {
```
miyehn, line 764: Looks like all usages of this function are given the time parameter, so we can make time a required param and get rid of the check for undefined

src/Game/Potency.ts
```diff
@@ -1,67 +1,71 @@
...
+	AF3, AF2, AF1, UI3, UI2, UI1, ENO, POT, PARTY
 }
 
 export type PotencyModifier = {
-	factor: number,
-	source: PotencyModifierType
+	source: PotencyModifierType,
+	buffType?: BuffType,
+	damageFactor: number,
+	critFactor?: number,
```
miyehn, line 13: We can prob make critFactor and dhFactor required (because it's not obvious what undefined mean in this context, but having them at 0 is very clear)

src/Game/Potency.ts
```diff
@@ -123,4 +144,42 @@ export class Potency {
...
 	hasSnapshotted() { return this.snapshotTime !== undefined; }
+
+	#calculatePotencyModifier(damageBonus?: number, critBonus?: number, dhBonus?: number) {
+		const base = this.#calculateDamage(controller.gameConfig.criticalHit, controller.gameConfig.directHit);
+		const buffed = this.#calculateDamage(controller.gameConfig.criticalHit, controller.gameConfig.directHit, damageBonus, critBonus, dhBonus);
+
+		return buffed / base;
+	}
+
+	#calculateDamage(crit: number, dh: number, damageBonus?: number, critBonus?: number, dhBonus?: number) {
```
miyehn, line 155: * make all params required to get rid of the undefined checks? With default values maybe
* rename `damageBonus` to `damageMultiplier` or `damageFactor`? Because it's multiplicative and defaults to 1 (aka it's not obvious that 1 means "no bonus")

miyehn (commented):
(more WIP)
All are super nit picky comments on details that don't affect functionality at all... they are more like notes to self, because I'm still assuming myself as the maintainer and might need to jump in at any time.
That said, they don't have to be addressed to be merged, unless you agree with them xD

src/Game/GameState.ts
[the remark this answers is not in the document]
TMTurtle: Agreed

src/Game/Buffs.ts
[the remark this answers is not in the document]
TMTurtle: buffConstants I think had a reason to exist at some point but doesn't need to anymore. I like that Buff holds the implementation of the mapping but its only used in two places so I could go either way on it.

src/Game/Common.ts
```diff
@@ -6,6 +6,18 @@ export const Debug = {
 	constantSlidecastWindow: true,
 };
 
+export const enum MarkerColor {
```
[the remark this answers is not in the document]
TMTurtle: Done

src/Game/Potency.ts
TMTurtle: Done

src/Game/Potency.ts
TMTurtle: Done. I'd originally had damageBonus as 0-based but it made more sense as a factor and it just never got renamed.

miyehn:
(more WIP...)

Potentially a bug? Currently pending damage doesn't take party buff into account yet, for example the fire 1 pending damage and the actual damage are different:
image
image

Oh just noticed the same thing happens for selection summary too:
image
image

At some point it might also be useful if we show the breakdown of partybuff potency calculation somewhere, like the below screenshot. Even if nobody uses it it could still help us debug :0 It wasn't done for pot because pot potency buff was very straightforward to calculate. Probably not for this PR but anyway I'm putting it into my bucket list of feature ideas...
image

TMTurtle:
image
Its an easy enough change to reflect buffs in the total. Do we need to change the UI to reflect that they're being included or is it implied by the word total and the summary breakdown?

A per-buff modifier, when not just a flat +5%, is possible to calculate in the same way we calculate the aggregate party buff contribution - we calculate a baseline total damage modifier, then the damage modifier with the additional buff contribution, then divide them. The interesting thing is that buffs that aren't flat damage modifiers aren't _far off_ from just being flat +5% damage modifier. As an example, at 2000 crit, Battle Litany calculates out to +4.734%, at 2500 +4.949%.

miyehn:
> image Its an easy enough change to reflect buffs in the total. Do we need to change the UI to reflect that they're being included or is it implied by the word total and the summary breakdown?
> 
> A per-buff modifier, when not just a flat +5%, is possible to calculate in the same way we calculate the aggregate party buff contribution - we calculate a baseline total damage modifier, then the damage modifier with the additional buff contribution, then divide them. The interesting thing is that buffs that aren't flat damage modifiers aren't _far off_ from just being flat +5% damage modifier. As an example, at 2000 crit, Battle Litany calculates out to +4.734%, at 2500 +4.949%.

Ah that's easy to change :)) Because party buff damage can't be toggled off (which could be another QoL change), I personally think it's implied that pending damage includes them. But if there's confusion we can also make a tooltip questionmark and explain it.

And now that I think about per-buff modifier again.. hmm also actually it might be hard to display party buff breakdown in the stats table, because the table groups skills by name but every instance of a skill may have different buffs. So if we want to display the breakdown somehow, maybe timeline is a better place..? And I'm not sure if it's better to display just percentage of damage increase, or amount or percentage of stats increase (CH/DH), or both...

Anyway, I'm totally just thinking out loud so don't worry about these QoL features. They prob still need more planning.

(I took a break and did some MSQ today haha, will prob finishing looking at the last few files this weekend)

PR #8, opened 2024-07-31
7.05 - Remove Paradox mp cost under ice

TMTurtle:


miyehn:
Ripp I fixed it before I saw this PR.. Thx for the help!

PR #13, opened 2024-08-18
Refactor default GameConfig and add consts for discord links

zqsz-xiv:
It seems like the actual default values were previously being set in Controller's constructor despite appearing several places throughout the codebase; this refactor unifies everything and makes sure there's only one place to change the default config.

`initialResourceOverrides` is deliberately left out of `DEFAULT_CONFIG` because I'm concerned that `DEFAULT_CONFIG.initialResourceOverrides` would share the same mutable array object. I don't think this will cause issues, but it's worth keeping in mind.


miyehn:
Ayyee thank you for cleaning up some ancient relics...

PR #14, opened 2024-08-18
Run eslint --fix

zqsz-xiv:
This fixes a bunch of warnings that were raised during `npm start`/`npm build` by running `./node_modules/.bin/eslint --fix --ext .js,.jsx,.ts,.tsx` src/`.

Notably, it removes a lot of calls to bind (which [don't work on arrow functions](https://eslint.org/docs/latest/rules/no-extra-bind)) and a couple trivial constructors ([one is generated by default](https://eslint.org/docs/latest/rules/no-useless-constructor)).

I removed a few unused imports, but there's still some variables the linter is reporting as unused that I elected not to touch for now. @miyehn when you get the time, can you check the remaining warnings and either suppress them or remove variables as needed?

### DIFF
--- a/src/Components/ColorTheme.tsx
+++ b/src/Components/ColorTheme.tsx
@@ -193,11 +193,11 @@ export class SelectColorTheme extends React.Component {
 	}
 
 	componentDidMount() {
-		getCurrentColorTheme = (()=>{return this.state.colorTheme}).bind(this);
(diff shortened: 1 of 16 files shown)


zqsz-xiv:
(also ideally we have a precommit hook or at least an `npm lint` command to run this automatically, but I think this is enough for now)

PR #15, opened 2024-08-21
FPS tax, shellVersion, and some cleanups that hopefully didn't break anything

miyehn:
It should have all the functionality there I just need to fill in text + update change log.
But I'll let it sit a day and check again tomorrow
If anyone has the time to also sanity check for me that'd be great 😂 (esp because you might still have some old files in your cache)


src/Game/GameConfig.ts
```diff
@@ -123,18 +135,29 @@ export class GameConfig {
 		}
 	}
 
+	getCasterPlusFpsTax(capturedCastTime: number) {
+		if (this.shellVersion < ShellVersion.FpsTax) {
+			return this.legacy_casterTax;
+		}
+		return (Math.floor(capturedCastTime * this.fps + 1) + Math.floor(0.1 * this.fps + 1)) / this.fps
```
zqsz-xiv, line 142: should we make the 0.1 a fixed "caster tax" constant somewhere?

src/Components/PlaybackControl.js
```diff
@@ -239,7 +247,8 @@ export class Config extends React.Component {
 					criticalHit: this.state.criticalHit,
 					directHit: this.state.directHit,
 					animationLock: this.state.animationLock,
-					casterTax: this.state.casterTax,
+					fps: this.state.fps,
```
zqsz-xiv, line 250: would it be possible to add an element (like a grayed-out, non-input number field) that displays the estimated tax on a 2.5/2.8/3.0s casts with the current config value? that would give the user some more feedback on what happens when a change is made

zqsz-xiv (commented):
left some comments on code/UI, I probably won't have time to test this against existing timelines in the next couple days but I should be able to run some tests closer to the weekend

a couple higher-level questions (we can discuss on discord for more visibility)
- what happens if you try to import an existing timeline? does it just populate `legacy_casterTax` and proceed accordingly? is there any actions a user can take to automatically convert a legacy timeline to the new calculation method?
- i don't think we would need to update calculations very frequently since the game's formulas are pretty stable and as far as i know, FPS tax is the last major thing we're missing, but i think we should try and make a more robust system for upgrading timelines between `ShellVersion`s. we could maintain a file of functions with signature `(GameRecord) -> GameRecord`, and each function describes how to update a timeline between versions. for example, the function to upgrade from version `0 -> 1` would modify the fields of the `GameConfig` object, and adjust the `waitTime`s between skills to align with the new caster tax formula. while this would require a bit more effort to maintain, it gives us more flexibility in potential future config/calculation/data format changes, and would let us silently migrate users across versions without worrying about invalidating/deleting their existing files.

src/Components/PlaybackControl.js
miyehn: That's a good point I should put it somewehre

zqsz-xiv (approved):
yippee

PR #18, opened 2024-08-27
Level sync

TMTurtle:
* Adds a level sync for 70, 80, 90, 100.
* Establishes a pattern for traits and level dependent skills
* Damage and Speed calculations respect downsynced stat formulas
* Needs a few translations
* Should double check the refactor around moving caster / fps tax to new class


miyehn:
o7 Ty Turtle! Now it looks like I'll be able to get to this starting next Monday-

src/Components/StatusDisplay.js
```diff
@@ -421,15 +434,17 @@ export class StatusDisplay extends React.Component {
...
 		}
 		updateStatusDisplay = ((newData)=>{
 			this.setState({
 				time: newData.time,
 				resources: newData.resources,
 				resourceLocks: newData.resourceLocks,
 				selfBuffs: newData.selfBuffs,
-				enemyBuffs: newData.enemyBuffs
+				enemyBuffs: newData.enemyBuffs,
+				gameState: newData.gameState,
```
miyehn, line 447: I think we should refrain from passing the entire `gameState` object into view components, because `gameState` exposes a lot of unnecessary information and adds to the risk of, say, accidentally modifying something in the gameState, or holding a dangling pointer to an outdated gameState. It also makes it a bit harder to read/debug code because it's no longer immiediately obvious what each each view component is responsible for rendering (because they have access to everything)
I haven't checked all usages of `gameState` but looks like a lot of them are just calls to `hasUnlockedTrait`, so we should be able to get around by passing in just `gameState.config.level`, and move the implementation of `hasUnlockedTrait` to somewhere outside of `GameState`, such as `Game/Traits.ts`. (Also see my other comment around `gameState.traitList`)

I think ideally we want to keep data flow as one-way as possible, which means:
- views get the bare minimum information to be able to render
- prefer taking values/deep copies over pointers

I have definitely taken a lot of shortcuts myself... for example passing in the `resources` map.... But we should still try to be somewhat vigilant, when it's easy enough to avoid xD

src/Game/GameState.ts
```diff
@@ -26,6 +27,8 @@ export class GameState {
 	cooldowns: CoolDownState;
 	eventsQueue: Event[];
 	skillsList: SkillsList;
+	displayedSkills: DisplayedSkills;
+	traitsList: TraitsList;
```
miyehn, line 31: `traitList`: I think we should move it out of the `GameState` class or at least make it static because it's the same for all `GameState` instances.

`displayedSkills`: I'm 80% inclined to also move it out of the `GameState` class. The only information it needs is `config.level`, which can probably be passed in as a function argument when querying

`skillsList`: now that I think about it, I'm also 60% inclined to move it out of `GameState`... I might try to do this after the pending PRs 🤔 

And the result will be a much smaller `GameState` class which costs less to create and toss away

miyehn (commented):
[WIP]
Currently, when the user has records in localStorage, they don't seem to be loaded correctly and result in NaNs. I haven't checked details in code but my guess is because levelsync is a new config we need to default it to 100 if it's not found in the cached record :0
image

TMTurtle:
NaN is actually coming from the switch cases:
image
This, for some reason, is falling through to the default case and returning undefined, which typescript doesn't think should be possible. I had liked this structure because it throws a compilation error if you add a new level without adding the new data for it - oh well.

miyehn:
* NaNs (one more change, or I can do it tomorrow if you're busy): [shanzhe's change](https://github.com/miyehn/ffxiv-blm-rotation/commit/16f4c94f5416afe41c7f18db5cbabb985404932d) does fix the NaN issue and lets number enums work with switch statements as expected, so let's go with that.
* Functionality (good to go): I didn't test super extensively but tried a ~2min section of Akairyu's file on lvl 90 and works fine and damage matches too.
* UI (good to go): I'm not sure why this misalignment happens but looks like it's triggering a bug in my ancient hacky code 😂 I'll debug once it's merged
image
* Code (good to go): there are still thing's I'm thinking about tweaking (`displayedSkills`, `skillsList` etc.) but they are just like any other refactor that never ends... let's hold them for now and do them at another time (if ever)

I'll fill in the "lvl" translation in Chinese. Again tysm!!! 🎉

miyehn:
Hrrrnnn the sanity test is failing because it doesn't recognize ResizeObserver.. I updated on main branch but know too little about tests to fix it on this pr... meh will deal with it later :/

src/Components/Localization.tsx
```diff
@@ -74,9 +77,11 @@ const skillsZh = new Map<SkillName, string>([
 const skillsJa = new Map<SkillName, string>([
 	[SkillName.Fire, "ファイア"],
 	[SkillName.Blizzard, "ブリザド"],
+	[SkillName.Fire2, ""],	// TODO
+	[SkillName.Blizzard2, ""], // TODO
```
TMTurtle, line 81: Need translations here as well

src/Components/Localization.tsx
```diff
@@ -74,9 +77,11 @@ const skillsZh = new Map<SkillName, string>([
...
 	[SkillName.Fire4, "ファイジャ"],
 	[SkillName.Transpose, "トランス"],
-	[SkillName.HighThunder, "ハイサンダー"],
+	[SkillName.Thunder3, ""], // TODO
```
TMTurtle, line 84: and here

miyehn:
Merging this now! In the future we don't need to worry about Japanese translation because it's not officially supported (yet) anyway 👌

PR #20, opened 2024-09-01
Implement image export feature

zqsz-xiv:
This PR gives users the ability to export a timeline, or a selection of a timeline, as an image. This is intended to help guide-makers and people who like to export images of timelines for easy reference.

Image export settings are found below the fight record import/export options:
image

Example (manual screenshot):
image

Exporting the selected spells (despair through the first F4):
fight

Example of exporting a big timeline and line wrapping every 20 seconds:
fight (1)


miyehn (commented):
[wip]

src/Components/ImageExport.tsx
```diff
@@ -0,0 +1,206 @@
...
+				filename={"fight"}
+				fileFormat={FileFormat.Png}
+				getContentFn={this.createMockCanvas.bind(this)}
+				displayName={localize({
+					en: "export selection as png",
+					zh: "将选择部分导出为png"
+				})}
+			/>
+			<Expandable
+				title={"Settings"}
```
miyehn, line 200: The title prop of `Expandable` is also used as id for caching its expand/collapse state, so we need to make sure it's unique (something more specific will do, like `PngExportSettings`). This is why title is always required even if `titleNode` exists. Sorry I didn't mention this anywhere

miyehn (commented):
This finishes my code comments.

Functionality-wise, I think right now it's good to go. There are a few things I would like to change but they are just like the other gazillion things I want to work on but never get to... So I think we can just keep them in the backlog for now:
* Currently, the exported timeline png has a sizable gap at its bottom (space below the green gcd bar) that I think we should eventually get rid of. Or make the margin size(s) configurable. But to do that I need to first put all the magic number definitions into one place so there's a reliable way of getting rendered elems dimensions...
* After ^this it'll also be easier to give options to export pngs at different resolutions
* Swapping out the 2d context is still a bit hacky, but before that there's the even hackier swapping out `controller.game` all the time... ouch
* The scrollable section to the right of main input (and some other places of the tool) is getting a bit cluttered, and I kinda want to start using tabs at some point. In this case the scrollable region would probably become tabs "current config", "edit config", "export", "control". And/or move export-related stuff to somewhere else

Tysm shanzhe!! I'll fill in the rest of Chinese translation tomorrow (I mean today).

src/Controller/Controller.ts
```diff
@@ -1329,5 +1338,15 @@ class Controller {
 			}
 		}
 	}
+
+	setImageExportConfig(newConfig: ImageExportConfig) {
+		this.imageExportConfig = newConfig;
+	}
+
+	generateTimelineImageURL(): string {
```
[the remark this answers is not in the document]
zqsz-xiv: yep, looks like I left it in by mistake

src/Components/ImageExport.tsx
zqsz-xiv: np, fixed

zqsz-xiv:
> Currently, the exported timeline png has a sizable gap at its bottom (space below the green gcd bar) that I think we should eventually get rid of.

If you want I can look into this before merging, it should be possible to do without changing any of the rendering constants (I can probably just resize the "fake" canvas objects).

>The scrollable section to the right of main input (and some other places of the tool) is getting a bit cluttered, and I kinda want to start using tabs at some point.

Definitely agree with this, even for more than just the control panels. Some simple stuff (like adding timeline markers) is placed in kind of an out-of-the-way spot.

miyehn:
> > Currently, the exported timeline png has a sizable gap at its bottom (space below the green gcd bar) that I think we should eventually get rid of.
> 
> If you want I can look into this before merging, it should be possible to do without changing any of the rendering constants (I can probably just resize the "fake" canvas objects).

I think it's not worth doing anything hacky for now because we can easily make it better when we cleanup the constants (and maybe also with configurable gap size + option to export at, say, 1.5x or 2x resolution)

> Some simple stuff (like adding timeline markers) is placed in kind of an out-of-the-way spot.

Yup... everything below the timeline is what I want to rework the most

PR #25, opened 2024-09-21
Job state generalization refactors (2/?)

zqsz-xiv:
This PR creates `BLMState`, a child class of `GameState`. BLM-specific skills and state have been adjusted to match this new format.

All skills now have their effects on the game state encoded in the `onConfirm` and `onApplication` functions. These should be specified either at the skill's declaration site, or for logic shared across many spells, added to a common helper function (`makeGCD_BLM`). While a particular skill's side effects can still be expressed in two different places (the declaration site and the body of `makeGCD_BLM`), it should now be considerably easier to find the logic than before.

Many frontend elements and the controller still make calls to BLM-specific components, like `state.getIceStacks()`, in order to properly render job gauges. For the time being, I've left those in with explicit casts from `GameState` -> `BLMState`, which I hope to refactor out in the near future.

Remaining tasks before we're truly job agnostic:
- Consolidate resource/CD length declaration (PCT still has overrides disabled)
- Remove casts from controller/frontend references, and figure out how to split job-specific logic in the frontend
- Refactor `getSkillAvailabilityStatus` and state-dependent button replacements (paradox, retrace)

### DIFF
--- a/src/Game/Resources.ts
+++ b/src/Game/Resources.ts
@@ -1,4 +1,5 @@
 import {Debug, ResourceType} from "./Common"
+import {EffectFn} from "./Skills";
 import {PlayerState, GameState} from "./GameState";
 import {ActionNode} from "../Controller/Record";
 
@@ -8,16 +9,16 @@ export enum EventTag {
 	LucidTick
 }
 
-export class Event<T extends PlayerState> {
+export class Event {
 	name: string;
 	#tags: EventTag[];
 	timeTillEvent: number;
 	delay: number;
-	effectFn: (state: T, node?: ActionNode) => void;
+	effectFn: EffectFn<PlayerState>;
 	canceled: boolean;
 
 	// effectFn : () -> ()
-	constructor(name: string, delay: number, effectFn: ()=>void) {
+	constructor(name: string, delay: number, effectFn: EffectFn<PlayerState>) {
 		this.name = name;
 		this.#tags = [];
 		this.timeTillEvent = delay;
--- a/src/Game/Resources.ts
+++ b/src/Game/Resources.ts
@@ -198,7 +198,7 @@ export class ResourceState {
 		let rsc = this.#map.get(rscType);
 		if (rsc) return rsc;
 		else {
-			console.assert(false);
+			console.assert(`could not find resource ${rscType}`);
 			return new Resource(ResourceType.Never, 0, 0);
 		}
 	}
--- a/src/Game/Resources.ts
+++ b/src/Game/Resources.ts
@@ -1,5 +1,5 @@
 import {Debug, ResourceType} from "./Common"
-import {PlayerState, GameState} from "./GameState";
+import {GameState} from "./GameState";
 import {ActionNode} from "../Controller/Record";
 import {BLMState} from "./Jobs/BLM";
 
--- a/src/Game/Resources.ts
+++ b/src/Game/Resources.ts
@@ -1,5 +1,4 @@
 import {Debug, ResourceType} from "./Common"
-import {EffectFn} from "./Skills";
 import {PlayerState, GameState} from "./GameState";
 import {ActionNode} from "../Controller/Record";
 import {BLMState} from "./Jobs/BLM";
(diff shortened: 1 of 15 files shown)


zqsz-xiv:
I noticed while fixing merge conflicts that attempting to load an invalid track (for example, loading a PCT file into the BLM site) will render the site unusable until a refresh. I'll do some more digging tomorrow to see if I can fix this.

zqsz-xiv:
> I noticed while fixing merge conflicts that attempting to load an invalid track (for example, loading a PCT file into the BLM site) will render the site unusable until a refresh. I'll do some more digging tomorrow to see if I can fix this.

Fixed, I forgot to make invalid skills always indicate that their requirements are not met.

src/Game/Skills.ts
```diff
@@ -886,13 +346,13 @@ export class SkillsList extends Map<SkillName, Skill> {
...
 		// TODO move contextual hotbar info (paradox, retrace) to here
 		const hotbarExcludeSkills = [
 			SkillName.Paradox,
 			SkillName.Retrace
 		];
-		for (const skillInfo of skillInfosMap.get(ShellInfo.job)!.values()) {
+		for (const skillInfo of skillMap.get(ShellInfo.job)!.values()) {
 			// Leave off abilities that are above the current level sync.
 			// Also leave off any abilities that auto-downgrade, like HF2/HB2/HT,
 			// since their downgrade versions will already be on the hotbar.
```
miyehn, line 358: rip we should exclude `SkillName.Never` as well:
image

src/Game/Skills.ts
```diff
@@ -1,883 +1,343 @@
...
+ *   // castTimeFn, recastTimeFn, etc. are valid here
+ * }
+ */
+export type Skill<T extends PlayerState> = Spell<T> | Weaponskill<T> | Ability<T>;
+
+// Map tracking skills for each job.
+// This is automatically populated by the makeWeaponskill, makeSpell, and makeAbility helper functions.
+// Unfortunately, I [sz] don't really know of a good way to encode the relationship between
+// the ShellJob and Skill<T>, so we'll just have to live with performing casts at certain locations.
+export const skillMap: Map<ShellJob, Map<SkillName, Skill<PlayerState>>> = new Map();
```
miyehn, line 133: If casting is inevitable would it be better if we wrap those casts into a helper function call, so they are kept in one place? So instead of exporting `skillMap`, we make it private to the file and instead export things like
```javascript
export function getSkill(job: ShellJob, skillName: SkillName) {
    // cast an return
}
export function getAllSkills(job: ShellJob) {
    // cast and return
}
```

miyehn (approved):
I think the structural changes all make sense (like... wow). I'm having a few more ideas on top of these but I should wait for you to finish first. At this point I'm not going to scrutinize every line of change and I think review is more about me catching up to the point that I can also jump in whenever.

TY for the heavy lifting 🙏

src/Game/Skills.ts
miyehn: Edit: oh I only saw a part of the problem. There's casting from the potentially undefined result from `Map.get`, and the casting between `Skill<PlayerState>` and `Skill<*State>`
I don't remember how it was done before tbh but wrapping them should at least put all the former (exclamation marks) into one place..?

src/Game/Resources.ts
```diff
@@ -197,7 +198,7 @@ export class ResourceState {
 		let rsc = this.#map.get(rscType);
 		if (rsc) return rsc;
 		else {
-			console.assert(false);
+			console.assert(`could not find resource ${rscType}`);
```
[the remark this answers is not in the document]
zqsz-xiv: whoops, actually meant for this to be `console.error`

src/Game/Skills.ts
zqsz-xiv: fixed

src/Game/Skills.ts
zqsz-xiv: I added `getSkill` and `getAllSkills` like you suggested. It seems like explicit casts aren't actually necessary, which seems a little unsound to me...? I don't think Typescript would be able to stop us from doing something like `const blmSkill: Skill<BLMState> = getSkill(ShellJob.PCT, SkillName.RainbowDrip)`, but it's at least useful for documentation.

The pattern I have in mind is [associated types](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html), where we'd be able to encode the relationship between a `ShellJob` and the corresponding `PlayerState` sub-class. Not sure if Typescript has a more idiomatic equivalent.

miyehn: Just tried and seems like the cast wasn't necessary because function return type is already explicit. So maybe it's fine...

zqsz-xiv: Yeah, I think the casts when populating the map in `makeAbility` and `makeSpell` are what's breaking the type system guarantees here. I don't think it's a big deal though, at least for the time being.

Maybe the better choice would just be to make `skillsMap` a separate property of each State class so there's no risk of any mixups.

miyehn: Yeah we can change any time 👌

PR #26, opened 2024-09-21
Avoid rendering during timeline replay

zqsz-xiv:
This PR stops display rendering logic from running during `loadBattleRecordFromFile`, significantly improving page load + timeline import time. `loadSlot` is still unnecessarily called twice in `tryAutoLoad`, but the added latency is no longer super meaningful.

Before/after (n=1, not scientific)
- Page load (with Akairyu's [7.0 TOP timeline](https://discord.com/channels/277897135515762698/1255782490862387360/1284925798364745771) as the active slot): 2150 ms -> 873 ms
- Import timeline from file (again using Akairyu's TOP timeline): 2478 ms -> 108 ms


miyehn (approved):
Wow... 👍

PR #28, opened 2024-09-26
Add hotbar icon for multi-stack abilities

zqsz-xiv:
Apparently, the number of stacks is rendered in-game as a font, and there's no sprite (I found a few similar-looking numbers in xivapi but there was no icon for 0). Instead, I picked the browser-default sans serif font and changed the colors accordingly:
- When the number of stacks is not zero, the number is displayed in white text w/ red border.
- When the number of stacks is zero, the number is displayed in red text w/ black border.

I made the drop shadow more intense if the ability is at maximum stacks to contrast better with the non-grayed out ability.

Here's what it looks like for picto (already deployed, unfortunately Pom Muse just has bad contrast with the skininess of the Windows default font)
image


miyehn (approved):
👌I'll come back and adjust its look later

PR #31, opened 2024-09-27
(1/?) Reorganize page layout

miyehn:
I'm using this PR as a space to dump thoughts, notes, screenshots etc. Comments and suggestions welcome.

Tabs might be useful. Currently bootstrapping some ideas:
image
image

TODO:
- [x] how to use tabs for the config section
- [x] figure out the new content hierarchy when/if using tabs
- [ ] the bottom of the page is so wide. Should probably use multiple columns to use screen space efficiently. Check: https://react-resizable-panels.vercel.app/ or https://ui.shadcn.com/

### DIFF
--- a/src/Components/Timeline.tsx
+++ b/src/Components/Timeline.tsx
@@ -1,4 +1,6 @@
-import React from 'react'
+/* eslint-disable @typescript-eslint/no-unused-vars */
+
+import React, {CSSProperties} from 'react'
 import {controller} from "../Controller/Controller";
 import {Help, Input, Slider} from "./Common";
 import {TimelineMarkerPresets} from "./TimelineMarkerPresets";
@@ -13,12 +15,18 @@ import {localize} from "./Localization";
 import {getCurrentThemeColors} from "./ColorTheme";
 import {getCachedValue, setCachedValue} from "../Controller/Common";
 
-export let updateTimelineView = () => {};
+import { LiaWindowMinimize } from "react-icons/lia";
+import { FaWindowMinimize } from "react-icons/fa6";
 
-export let scrollTimelineTo = (positionX: number)=>{}
+export let updateTimelineView = () => {
+};
+
+export let scrollTimelineTo = (positionX: number) => {
+}
 
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-let getVisibleRangeX = () => {}
+let getVisibleRangeX = () => {
+}
 
 // the actual timeline canvas
 class TimelineMain extends React.Component {
@@ -31,6 +39,7 @@ class TimelineMain extends React.Component {
 		visibleWidth: number,
 		version: number
 	};
+
 	constructor(props: {}) {
 		super(props);
 		this.state = {
@@ -42,7 +51,7 @@ class TimelineMain extends React.Component {
 		}
 		this.myRef = React.createRef();
 
-		this.updateVisibleRange = (()=>{
+		this.updateVisibleRange = (() => {
 			if (this.myRef.current) {
 				this.setState({
 					visibleLeft: this.myRef.current.scrollLeft,
@@ -51,6 +60,7 @@ class TimelineMain extends React.Component {
 			}
 		});
 	}
+
 	componentDidMount() {
 		this.setState({
 			timelineWidth: controller.timeline.getCanvasWidth(),
@@ -64,7 +74,7 @@ class TimelineMain extends React.Component {
 			});
 		});
 
-		scrollTimelineTo = ((positionX: number)=>{
+		scrollTimelineTo = ((positionX: number) => {
 			if (this.myRef.current != null) {
 				let clientWidth = this.myRef.current.clientWidth;
 				this.myRef.current.scrollLeft = positionX - clientWidth * 0.6;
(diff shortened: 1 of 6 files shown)


miyehn:
Content hierarchy thoughts: the two sections below will probably become tabs, as shown in screenshots above

### (to the right of input aka "props related to the main input window")
Overview (result of config and displayed in a form that would be directly helpful for viewing things)
* level
* displayed GCD, GCD after fps tax + any manual correction
* cast times table (if caster job)
* lucid/dot tick offsets
* CH, DH, CDH rates?
* proc mode
* resource overrides

Edit Config
* edit section + resource overrides
* "apply and reset"

Control
* auto pause / manual

Import/Export (should this section stay here?)
* Import
  * txt (file selector)
* Export
  * txt
  * txt, saving current game state as resource overrides (if this ever gets implemented). When loaded, the timeline will be empty but will have a bunch of resource overrides applied
  * csv for visualization
  * csv for Amarantine's combat sim (if it gets implemented)

### (bottom of page, aka "settings related to the timeline visualization"). Also each tab will have a fixed height. Use scrollbars and/or multiple columns if content overflow
Display settings
* move "tincture potency buff %" and "exclude damage when untargetable" to the stats section? because they affect stats output but not timeline visualization
* can have checkboxes for which element(s) to display: MP ticks, damage marks, etc. Maybe other types of visual elements as well
  * thinking about displaying dot damage using accent color instead
* maybe put the "save image" button and related settings here? Or put it (and all other import/exports such as damage log csv) into the import/export tab on the right of main input?

Timeline Markers

Timeline Editor

---

Main questions for now:
* should all imports and exports stay together, or stay at where they are most relevant? Mostly talking about damage log csv and image export and if they should stay close to other txt/csv exports. Skill seq presets and marker imports should probably stay where they are.
* should the import/export section (especially if it keeps getting larger) be moved to somewhere else, such as below "skill sequence presets"?

zqsz-xiv:
> lucid/dot tick offsets

These aren't currently configurable, so this would be a new feature, right?

> can have checkboxes for which element(s) to display: MP ticks, damage marks, etc. Maybe other types of visual elements as well

Looks like my hacks for image export will be useful elsewhere... though we might want to refactor them a bit.

---

> should all imports and exports stay together, or stay at where they are most relevant? Mostly talking about damage log csv and image export and if they should stay close to other txt/csv exports. Skill seq presets and marker imports should probably stay where they are.

I think all the timeline import/export buttons should stay grouped together. It might also be worth considering just putting the input/export tabs in both the config and visualization tab sets since it arguably makes sense as both "thing related to inputs" and "thing related to visualization." If we had to pick one, my personal preference would be at the bottom with the settings just because the tab name would be more noticable.

> should the import/export section (especially if it keeps getting larger) be moved to somewhere else, such as below "skill sequence presets"?

I think having its own settings tab (as long as it's noticable) should be fine. I find the "Skill sequence presets" section and surrounding text easy to miss.

miyehn:
> lucid/dot tick offsets

No they are not directly configurable. They are random depending on random seed. I just kept them in case people want that readonly info
image

> Looks like my hacks for image export will be useful elsewhere... though we might want to refactor them a bit.

Yeah if they are configurable for export I think there's no reason to not make it available on the site as well. I think now it looks ok but if we add any more visual element to the timeline it'll look too busy. And we do whatever refactor to keep it maintainable ofc

miyehn:
Version history is starting to get ugly I should merge this PR and open a new one :0
So far using tabs at the bottom and moved load/save to one of those tabs. Within each tab everything still needs to be organized. I'll work on those right after this.
image
image
image
image

zqsz-xiv:
I can take a look tomorrow evening if you want some more eyes on it. Otherwise feel free to merge if you want to work on other layout stuff.

src/Components/Timeline.tsx
```diff
@@ -103,7 +115,7 @@ class TimelineMain extends React.Component {
 				overflowX: "scroll",
 				overflowY: "clip",
 				outline: "1px solid " + getCurrentThemeColors().bgMediumContrast,
-				marginBottom: 10,
+				//marginBottom: 10,
```
zqsz-xiv, line 118: should we delete this?

zqsz-xiv (approved):
Nice work! A couple comments:
- I noticed on page load that the tabs start minimized by default. I think we should pick one to be open by default (leaving minimize optional), since text below the timeline can be easy to miss at first glance.
- I'd suggest re-ordering the tabs based on frequence of usage (from left to right: Timeline markers -> Timeline editor -> Import/Export -> Display settings) and making the first tab on by default.
- Can you make the headings for fight export bolded to match image export? The tab looks a bit visually uneven as-is.
image
- I don't think image export settings need to be in an expandable anymore, since there's plenty of space in the tab for all the options.

miyehn:
> I noticed on page load that the tabs start minimized by default. I think we should pick one to be open by default

I'm going to keep them collapsed by default for now because I'm more concerned otherwise:
* exposing too many functionalities can make it look daunting to first time users. There are already many comments about it seem daunting so I'm trying not to add to that. If people miss them on first glance it's fine. After they learn the basics (setting config and using skills), whenever they see these tab titles they'll probably just click around and figure out.
* people might not realize it can be collapsed and the whole section just always take up the bottom of the screen

Ideally I should observe how people actually learn to use it but nah I'm lazy... Lmk if you hear comments from people though (you don't count you know too much about the tool already)

> I'd suggest re-ordering the tabs based on frequence of usage

👌 

> make the headings for fight export bolded to match image export
I don't think image export settings need to be in an expandable anymore

Content layout within each tab is still to be adjusted, will probably come in the next PR. In fact I'll probably move the image export settings to under display settings, and move the current display settings (tincture buff % etc.) to somewhere in the stats section.

PR #34, opened 2024-10-01
add Japanese translation on Localization.tsx

totti:
I’ve decided to take small steps to avoid conflicts.


miyehn:
Yes that's a good idea 👍

PR #37, opened 2024-10-06
Show error timestamp in timeline editor

zqsz-xiv:
If moving/deleting a skill in the timeline editor invalidates some later action, the timeline editor will report the name of the offending skill in text, and display a little red arrow next to it in the timeline box. In situations where a change to an early portion of the timeline affects a skill usage far later (possibly multiple minutes away), it can be difficult to identify the issue. Adding an explicit timestamp to the error message helps alleviate this.

image

### DIFF
--- a/src/Components/TimelineEditor.tsx
+++ b/src/Components/TimelineEditor.tsx
@@ -176,8 +176,15 @@ export class TimelineEditor extends React.Component {
 							nodeName = node.skillName ?? "(unknown skill)";
 						}
 					}
+					let errorMessage = "This sequence contains invalid actions! Check: " + nodeName;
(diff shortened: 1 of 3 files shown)


src/Components/TimelineEditor.tsx
```diff
@@ -176,8 +176,15 @@ export class TimelineEditor extends React.Component {
 							nodeName = node.skillName ?? "(unknown skill)";
 						}
 					}
+					let errorMessage = "This sequence contains invalid actions! Check: " + nodeName;
```
miyehn, line 179: This works too, or there's also `StaticFn.displayTime` in `Common.tsx` if we want to use than existing function to conver time display format

miyehn (approved):
🙏

Issue #36, opened 2024-10-06
Reorganize page layout


miyehn:
Maybe issues is a better dumping ground than PRs when something takes longer and needs multiple PRs. Anyway this is a continuation of #31.

### TODO

Right
- [x] add "level" to Overview
- [x] line height for Overview and Control tabs
- [x] show exact overrides in the overview tab

Bottom

- [x] Timeline markers
  * Columns "presets", "load/save", "add"
- [x] Import/Export
  - [x] Columns "fight export", "fight import", "image export"
  - [x] move the image export checkboxes to display settings (and apply them for general timeline canvas rendering)
  - [x] adjust multi-line image export canvas height & gap between lines

Misc
- [x] figure out one of the resizeable column modules and make a common component for it (or for now just a generic columns wrapper)
- [x] update instructions if needed
- [x] localization

miyehn:
should be done after #42 gets in so closing this

PR #38, opened 2024-10-07
Reorganize page layout (2/?)

miyehn:
[grand todo list](https://github.com/xivintheshell/xivintheshell/issues/36#issue-2568536028)

90% of the way there for this batch of rework, I think. Laid out a lot of information in plain sight hopefully to save some clicks and navigation, but lmk if something just became more confusing...

Here are some screenshots:
image
image
image
image
image
image

"display settings" tab is hidden because there's nothing to show yet. Will probably move over the checkboxes for optional elements in the next PR.

### DIFF
--- a/src/Components/Common.tsx
+++ b/src/Components/Common.tsx
@@ -257,7 +257,7 @@ export function Tabs(props: {
 	useEffect(() => {
 		let selected: number | undefined = props.defaultSelectedIndex;
 		// if a cached value exists, it will always override the default
-		let cachedSelected = getCachedValue(props.uniqueName + "SelectedTab");
+		let cachedSelected = getCachedValue("tabs: " + props.uniqueName);
 		if (cachedSelected !== null) {
 			if (cachedSelected === "none") {
 				selected = undefined;
@@ -317,7 +317,7 @@ export function Tabs(props: {
 			if (!isSelectedTab) { newIndex = i; }
 			else if (props.collapsible) { newIndex = undefined; }
 			setSelectedIndex(newIndex);
-			setCachedValue(props.uniqueName + "SelectedTab", newIndex===undefined ? "none" : `${newIndex}`);
+			setCachedValue("tabs: " + props.uniqueName, newIndex===undefined ? "none" : `${newIndex}`);
 		}}>{props.content[i].titleNode}</span>);
 
 		content.push(<div key={i} style={{
@@ -328,7 +328,10 @@ export function Tabs(props: {
 	if (props.collapsible && selectedIndex !== undefined) {
 		titles.push(<span
 			key={titles.length}
-			onClick={() => {setSelectedIndex(undefined);}}
+			onClick={() => {
+				setSelectedIndex(undefined);
+				setCachedValue("tabs: " + props.uniqueName, "none");
+			}}
 		><LiaWindowMinimize style={{
 			marginLeft: 16,
 			fontSize: 14,
@@ -344,7 +347,7 @@ export function Tabs(props: {
 		<div>{titles}</div>
 		<div className={"staticScrollbar"} style={{
 			display: selectedIndex === undefined ? "none" : "block",
-			height: props.height - titleHeight,
+			height: props.height - TABS_TITLE_HEIGHT,
 			boxSizing: "border-box",
 			padding: "10px 5px",
 			overflowY: "scroll"
@@ -388,6 +391,7 @@ export class Input extends React.Component {
 }
 
 type SliderProps = {
+	uniqueName: string,
 	onChange?: (e: string) => void,
 	defaultValue?: string,
 	description?: ContentNode
@@ -398,6 +402,7 @@ type SliderState = {
 }
 export class Slider extends React.Component {
 	props: SliderProps = {
+		uniqueName: "anonSlider",
 		defaultValue: "default slider value",
 		description: "default description"
 	};
@@ -412,10 +417,19 @@ export class Slider extends React.Component {
 		this.onChange = ((e: ChangeEvent<{value: string}>)=>{
 			this.setState({value: e.target.value});
 			if (typeof this.props.onChange !== "undefined") this.props.onChange(e.target.value);
+			setCachedValue("slider: " + this.props.uniqueName, e.target.value)
 		});
 	}
 	componentDidMount() {
-		if (typeof this.props.onChange !== "undefined") this.props.onChange(this.state.value);
+		let initialValue = this.state.value;
+		let str = getCachedValue("slider: " + this.props.uniqueName);
+		if (str !== null) {
+			initialValue = str;
+			this.setState({value: initialValue});
+		}
+		if (typeof this.props.onChange !== "undefined") {
+			this.props.onChange(initialValue);
+		}
 	}
 	render() {
 		return <div style={{...{display: "inline-block"}, ...this.props.style}}>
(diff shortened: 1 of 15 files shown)


zqsz-xiv:
Some high-level thoughts before I look at the code:

- Pretty minor thing (I forgot to check this in the first tabs PR): the bottom tabs get pretty mangled on mobile devices; I don't think we need to make supporting phones a priority (since presumably most people do rotation planning on a computer of some kind) but I think enough people will try to throw together ad-hoc lines in a phone browser that we should at least try to make elements not collide.
image

- Not sure about splitting the "config" and "overview" panels. Being able to see your GCD value while adjusting sps is useful. I do think splitting out "control" into its own tab makes sense.

- When I tile BLM in the Shell to half the screen on my second monitor the columns look fine, but when I make the site take up the whole width of a monitor the distance between vertical columns feels a little too large (see screenshot). Instead of making the columns always take up a certain percentage of the screen, is it possible to make them just stay at a fixed width if the browser window is wider than some threshold? Other people might have different opinions on how this looks (basically whether you want the empty space to be on the end rather than in between), but I'd prefer leaving empty space at the end of the columns so information is more condensed and it's less disruptive if we add more columns in the future.
image

- Can you add back the vertical box boundaries for the timeline editor/"timeline is up to date" components? It feels a little jarring not seeing borders at the top/bottom.

Old (with boundaries):
image

New (without boundaries):
image

src/Components/Common.tsx
```diff
@@ -342,16 +347,54 @@ export function Tabs(props: {
...
+	children: {
+		content: ContentNode,
+		title?: ContentNode,
+		defaultSize?: number,
+		minSize?: number
+	}[]
+}) {
+	let colors = getCurrentThemeColors();
+	let children: React.ReactNode[] = [];
+	for (let i = 0; i < props.children.length; i++) {
```
zqsz-xiv, line 371: nit: IMO this would be a little easier to read as a  `props.children.map(...)` function + `push` the `PanelResizeHandle` at the very end

src/Components/Common.tsx
```diff
@@ -412,10 +457,19 @@ export class Slider extends React.Component {
 		this.onChange = ((e: ChangeEvent<{value: string}>)=>{
 			this.setState({value: e.target.value});
 			if (typeof this.props.onChange !== "undefined") this.props.onChange(e.target.value);
+			setCachedValue("slider: " + this.props.uniqueName, e.target.value)
```
zqsz-xiv, line 460:
```suggestion
			setCachedValue("slider: " + this.props.uniqueName, e.target.value);
```

src/Components/Common.tsx
```diff
@@ -412,10 +457,19 @@ export class Slider extends React.Component {
...
 	}
 	componentDidMount() {
-		if (typeof this.props.onChange !== "undefined") this.props.onChange(this.state.value);
+		let initialValue = this.state.value;
+		let str = getCachedValue("slider: " + this.props.uniqueName);
+		if (str !== null) {
+			initialValue = str;
+			this.setState({value: initialValue});
+		}
+		if (typeof this.props.onChange !== "undefined") {
```
zqsz-xiv, line 470: why does this need a `typeof` instead of a direct `=== undefined` check?

src/Components/DamageStatistics.tsx
```diff
@@ -180,6 +181,80 @@ function PotencyDisplay(props: {
...
+		// state
+		this.state = {
+			tinctureBuffPercentageStr: "8",
+		}
+
+		// tincture buff percentage
+		str = getCachedValue("tinctureBuffPercentage");
+		if (str !== null) {
+			this.state.tinctureBuffPercentageStr = str;
+		}
```
zqsz-xiv, line 209:
```suggestion
		// tincture buff percentage
		str = getCachedValue("tinctureBuffPercentage");

		// state
		this.state = {
			tinctureBuffPercentageStr: str ?? "8",
		}
```

src/Components/TimelineMarkers.tsx
```diff
@@ -296,132 +269,178 @@ export class TimelineMarkerPresets extends React.Component {
...
+					}
+					controller.timeline.addMarker(marker);
+					controller.updateStats();
+					e.preventDefault();
+				}}>{localize({en: "add marker", zh: "添加标记"})}
+			</button>
+		</form>;
+
+		// https://github.com/OverlayPlugin/cactbot/tree/main/ui/raidboss/data/07-dt
+		let presetsSection = <div style={{
```
zqsz-xiv, line 399: would it be better to keep this (or at least the list of displayName + urls for each fight) as a global variable so it's easier to update between tiers?

src/Components/Common.tsx
miyehn: I would've used map or forEach but I need something unique for each iteration to be part of the `key` prop 🤔..

zqsz-xiv: you can do `.map((obj, i) => {})` or `.forEach((obj, i) => )` to get the positional index during iteration

src/Components/Common.tsx
miyehn: I don't remember... this looks like some ancient code.. I can change it to `===`

miyehn:
Oh lmao I didn't expect you to look at the gazillion code changes tbh, most were moving stuff around and not very interesting :0

> mobile devices

Nooo.... for my own hobby projects, I'd have features that are either supported (with some standard including UI and aesthetics) or not supported, and no mobile device is not supported. I'm now tempted to hide everything and just show a static message saying "please visit from a desktop device" if the screen width is below some threshold, to make it clear for users.

But this is no longer 100% my personal project so uh.... no I still don't want to work on anything related to responsiveness, for mostly personal reasons. But I won't stop you or anyone else ;_;

>  splitting the "config" and "overview" panels

If it's just GCD you think is important when editing, it can be displayed as a preview number like caster tax. Or at most 5 lines of additional text maybe. I put quite some stuff in overview and unless I get rid of most of them or put things an expandable (which will also require an additional click, plus layout shift), combining the two tabs could make that section huge again... what do you think?

> columns too spread out on big monitors
> add back the vertical box boundaries for the timeline editor/"timeline is up to date" components

Valid and easy to fix 👌

src/Components/TimelineMarkers.tsx
miyehn: I'd say meh, this is far from being the messiest part of the code. And update between tiers that's like, 5 lines every 3 months anyway 🤔

zqsz-xiv:
> I still don't want to work on anything related to responsiveness on mobile, for mostly personal reasons. But I won't stop you or anyone else

I don't think we should explicitly disable the site on mobile, but yeah I agree it's fine to not design with it in mind. That said, if I have some time later I'll see if I can circle back and fix collisions like this (I do very occasionally use the site on my phone myself), but I won't make it a priority.

> I put quite some stuff in overview and unless I get rid of most of them or put things in an expandable (which will also require an additional click, plus layout shift), combining the two tabs could make that section huge again

Except the cast time table, the only information unique to the new "overview" tab is the cast time table (which I think is fine to keep in an expandable) and the lucid tick + thunder tick offsets (which are fairly small).

I don't think level + sps + GCD have value as read-only fields: even though they're more important than the other configuration fields, if I'm a user opening the "overview" tab and I see the level + sps + GCD text, I would very quickly ask myself "where would I go to change these?" values, and then have to move to the "config" tab to do so.

> If it's just GCD you think is important when editing, it can be displayed as a preview number like caster tax.

I think it makes sense to do this, so the place you see your GCD is the same place you edit it. I'd prefer removing the "overview" tab, and moving thunder + lucid offset back to "config" since they're small. Users would then see level + sps + GCD (the most important configuration options + maybe bold/recolor GCD for emphasis) right at the top of the "config" window where they can quickly edit them. I think the cast time table should be either kept as an expandable near the sps/GCD value, or made into a mouseover like the tax table: it's useful information, but IMO not as important as level + sps + GCD.

src/Components/Common.tsx
miyehn: I changed to a `forEach` loop instead of `map` because the resize handles have to stay between panels I believe so can't just push them all at the end

miyehn: ^actually... changed back to classic for loop because that loop now contains more logic...

miyehn:
All are valid arguments. Then I think I'll combine the two tabs and mostly restore what was there before + GCD preview next to sps (will do these tomorrow)

miyehn:
As of now:
image
image

miyehn:
Actually, just got rid of one more line:
image

zqsz-xiv:
Looks good! Just one more nitpick:
image

When expanded to fit my screen, the columns in the bottom tabs are pretty far offset from the tab selectors themselves. My personal preference would be for the tabs to keep the same width but be left-justified so my eyes don't have to move as far from the tabs to the actual behavior.

Not really sure what the best option here is; I'll defer to your judgement.

PR #39, opened 2024-10-10
Add ability to resize ability selections

zqsz-xiv:
shift-clicking an ability timeline now lets you resize the selection around the first chosen selection

<details><summary>demo</summary>

</details>

<details><summary>analogous behavior from google sheets</summary>

</details>

the orange timeline cursor shifting around to always match the start of the selection is a little weird (note that google sheets keeps a selection box on the initially-selected square), but I think it's reasonable to keep this behavior and it would be a bit more effort to change


miyehn (approved):
🚀

PR #40, opened 2024-10-11
Make resource display job-agnostic

zqsz-xiv:
(Mostly) separates out the logic for rendering job-specific gauge, buffs, and debuffs. The `GameState` base class now requires its children to define `getEnemyBuffViewProps(): BuffProps[]`, `getSelfBuffViewProps(): BuffProps[]`, and `getResourceViewProps(): ResourceDisplayProps[]`, which contain the properties necessary to render the corresponding components. Cutting out the intermediate layer of abstraction (`StatusResourcesViewProps`, `StatusEnemyBuffsViewProps`, `StatusSelfBuffsViewProps`) and pushing more work to the `GameState` class makes it easier to reason about types, since specializing these props for each job would have been a nightmare.

In the course of debugging this PR, I found that we have some implicit assumptions about webpack's module resolution order in our code: `Game/Jobs/BLM.ts` _must_ be initialized before `Components/Skills.tsx`, or else the `skillIcons` map will not be populated with BLM abilities. It should be possible to refactor some stuff up to allow implicit initialization of `skillIcons` while being relatively agnostic to import order, but I don't think it's worth the effort at the moment. Note that this import order issue is why I had to staple all the props computation to `GameState` + keep `buffIcons` population in `StatusDisplay.tsx`, as creating a new `Components/Jobs/BLM.tsx` file with all this logic, as I had initially planned, messed up the import order and cause a lot of icons to be ready prematurely.


src/Components/StatusDisplay.tsx
```diff
@@ -32,33 +18,43 @@ type StatusResourceLocksViewProps = {
...
-	thunderheadCountdown: number,
-	manawardCountdown: number,
-	swiftcastCountdown: number,
-	lucidDreamingCountdown: number,
-	surecastCountdown: number,
-	tinctureCountdown: number,
-	sprintCountdown: number
-}
+export interface ResourceBarProps {
+	kind: "bar";
```
miyehn, line 31: For now it's fine but I'll probably change `kind` to an enum once I get to make at least another common resource component to better display multiple resources of 1 stack- after I finish  messing with tabs

miyehn (approved):
I think this can go in for now, but it's definitely a TODO at some point to move the UI logic back to `Components/`. I locally messed with the `wip` commit a bit and think it should be doable. When/if any of us get to it I guess

src/Components/StatusDisplay.tsx
zqsz-xiv: It seems like the generally accepted way of implementing a discriminated union like this is to use a string literal on the discriminant field (https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) rather than an enum:

```
interface Circle {
  kind: "circle";
  radius: number;
}
 
interface Square {
  kind: "square";
  sideLength: number;
}
 
type Shape = Circle | Square;
```

The type checker will raise an error (though the message isn't always super clear) if you set "kind" to an invalid value. If we want to add another component type, we'd just declare a new interface and add it to the `ResourceDisplayTypes` union type.

zqsz-xiv:
I moved the new logic back into the Components folder, under `Components/Jobs/BLM.tsx`. It seems like the import order issue only happened when I tried to access this new file from within `Game/Jobs/BLM.ts` (I wanted to have each job class declare the corresponding component class).

PR #42, opened 2024-10-17
Move some image export settings to display settings tab

miyehn:
and maybe add a few more settings if it's convenient enough

11/2: I think it's ready:

Image export settings right now:
image

Display settings tab:
image

As a result it can export with everything
fight (59)

as well as with fewer elements, and line height adjusts accordingly:
fight (60)

### DIFF
--- a/src/Components/ImageExport.tsx
+++ b/src/Components/ImageExport.tsx
@@ -113,7 +113,9 @@ export class ImageExport extends React.Component<{}, ImageExportConfig> {
 				let currentHeight = 0;
 				if (this.state.includeTime) {
 					currentHeight += drawRuler(timelineOrigin, true);
-					currentHeight += drawMarkerTracks(timelineOrigin, currentHeight, true);
+					if (activeRenderProps.drawOptions.drawMarkers) {
+						currentHeight += drawMarkerTracks(timelineOrigin, currentHeight, true);
+					}
 				}
 				drawTimelines(timelineOrigin, currentHeight, true);
 			}
--- a/src/Components/ImageExport.tsx
+++ b/src/Components/ImageExport.tsx
@@ -60,7 +60,8 @@ export class ImageExport extends React.Component<{}, ImageExportConfig> {
 		if (startTime === endTime) {
 			// No selection was made, so export the whole timeline
 			startTime = -activeRenderProps.countdown;
-			endTime = controller.game.time;
+			const DRAW_EXTRA_DURATION = 4;// todo: ideally we'd want to have this just long enough to capture the last skill
+			endTime = controller.game.time - activeRenderProps.countdown + DRAW_EXTRA_DURATION;
 		}
 		const exportConfig = controller.imageExportConfig;
 		let nRows: number;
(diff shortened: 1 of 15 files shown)


src/Components/ImageExport.tsx
```diff
@@ -82,21 +83,20 @@ export class ImageExport extends React.Component<{}, ImageExportConfig> {
...
 		const dummyOneRowCanvas = document.createElement("canvas");
 		const dummySplitCanvas = document.createElement("canvas");
-		const ICON_SIZE = 28;
-		const ROW_PADDING = ICON_SIZE / 2;
+		const ROW_PADDING = TimelineDimensions.trackHeight / 2; // can make this a config if we want
 		const tlController = controller.timeline;
-		dummyOneRowCanvas.height = activeRenderProps.timelineHeight;
 		// We have no control over where to start drawing, so the one-row canvas should always
 		// have the whole duration
+		// note [myn]: iirc there's a hard limit for html canvas dimensions, so this might fail to capture very long
```
zqsz-xiv, line 90: Seems like it's 2^16 - 1 or 2^15 - 1 in most browsers: https://jhildenbiddle.github.io/canvas-size/#/?id=test-results

Hopefully that never becomes a concern!

src/Components/TimelineDisplaySettings.tsx
```diff
@@ -0,0 +1,28 @@
...
+			drawDamageMarks: drawDamageMarks,
+			drawMPTickMarks: drawMPTickMarks,
+			drawBuffIndicators: drawBuffIndicators,
+		};
+		controller.setTimelineOptions(options);
+	};
+	const [drawMarkers, setDrawMarkers] = useStateWithCallback<boolean>(true, setTimelineOptions);
+	const [drawDamageMarks, setDrawDamageMarks] = useStateWithCallback<boolean>(true, setTimelineOptions);
+	const [drawMPTickMarks, setDrawMPTickMarks] = useStateWithCallback<boolean>(true, setTimelineOptions);
+	const [drawBuffIndicators, setDrawBuffIndicators] = useStateWithCallback<boolean>(true, setTimelineOptions);
```
zqsz-xiv, line 20: Seems like `use-state-with-callback` was last updated 3 years ago, which makes me a little wary since my understanding of the JS/React system is that it tends to churn pretty quickly. I'm not opposed to adding new convenience dependencies but I think this one might be unnecessary.

Also, are the `draw*` state variables being read at all here? It seems like `Checkbox` reads from localStorage, so can you just use `setTimelineOptions({..., drawMarkers})` directly instead of `setDrawMarkers`?

src/Components/TimelineDisplaySettings.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+	return <div>
+		<Checkbox uniqueName={"drawMarkers"} label={localize({en: "draw timeline markers"})} onChange={setDrawMarkers}/>
```
zqsz-xiv, line 23: don't think this makes sense to include as its own display option, and it might confuse users if they uncheck this box, try to add timeline markers, and then get confused becaus they forgot about this setting

src/Components/TimelineDisplaySettings.tsx
```diff
@@ -0,0 +1,28 @@
...
+		<Checkbox uniqueName={"drawDamageMarks"} label={localize({en: "draw damage marks"})} onChange={setDrawDamageMarks}/>
+		<Checkbox uniqueName={"drawMPTickMarks"} label={localize({en: "draw MP and lucid ticks"})} onChange={setDrawMPTickMarks}/>
+		<Checkbox uniqueName={"drawBuffIndicators"} label={localize({en: "draw buff indicators"})} onChange={setDrawBuffIndicators}/>
```
zqsz-xiv, line 26: maybe change the label to "show damage marks" etc.

zqsz-xiv (changes requested):

zqsz-xiv:
wow my comments came from the future because of daylight savings time
image

src/Components/TimelineDisplaySettings.tsx
miyehn: Oop and I forgot to translate...

zqsz-xiv (approved):
👌

miyehn:
Oh wait you so fast, I also wanted to sneak in something small and was lazy to make another branch etc:
image

PR #44, opened 2024-11-03
Check MP cost at cast confirm window

zqsz-xiv:
If Enochian drops midway through a cast, the MP cost of a spell should be checked again at the cast confirm window, and the spell canceled if the MP cost is no longer met.

Two manual test timelines:
- [bad_b3_eno_drop.txt](https://github.com/user-attachments/files/17609177/bad_b3_eno_drop.txt)
(works on main, fails after this PR): this is also added as an automated test; after reaching 0 MP in a fire phase, the B3 cast was free under AF3 but costs 800 MP instead when eno drops, making the cast fail
- [b3_should_be_800_mp.txt](https://github.com/user-attachments/files/17609180/b3_should_be_800_mp.txt) B3 succeeds but consumes 800 MP after the eno drop happens

### DIFF
--- a/src/BlmRotations.test.ts
+++ b/src/BlmRotations.test.ts
@@ -469,3 +469,43 @@ it("replaces f1/b1 with paradox when needed", testWithConfig({level: 100}, () =>
 		},
 	]);
 }));
+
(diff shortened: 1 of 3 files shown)


miyehn (approved):
👌👌

PR #46, opened 2024-11-03
PCT merge cleanup (1/2)

zqsz-xiv:
3 changes:
- deleted StatsModifier.ts and moved logic into the BLMState class
- moved highlighting logic into a `highlightIf` field for each skill
- split job-specific colors into separate types


src/Game/Jobs/BLM.ts
```diff
@@ -126,6 +126,13 @@ export class BLMState extends GameState {
 		recurringPolyglotGain(this.resources.get(ResourceType.Polyglot));
 	}
 
+	override captureManaRegenAmount(): number {
```
zqsz-xiv, line 129: For some reason, if I didn't give this a `number` return annotation I was getting a ton of compiler errors along the lines of "type BLMState is not assignable to GameState," and the compiler wasn't nice enough to tell me the root cause...

Google seems to indicate something to do with the type of `this` being different between the child and parent classes and something about contravariance, but I don't know why that would cause typescript to freak out. I'm going to choose to not worry about it.

src/Game/Skills.ts
```diff
@@ -28,8 +28,9 @@ export type ConditionalSkillReplace<T extends PlayerState> = {
  * usage is confirmed.
  */
 export type ResourceCalculationFn<T> = (state: Readonly<T>) => number;
+export type StatePredicate<T> = (state: Readonly<T>) => boolean;
```
miyehn, line 31: Do you think we can replace `ValidateAttemptFn` and `IsInstantFn` with `StatePredicate`? To have 2 less types to remember. Unless you think these types may differ depending on how implementation for other jobs go :0

zqsz-xiv: done

PR #47, opened 2024-11-05
PCT merge cleanup - remove Hammer aspect and getPotencyModifiersFromResourceState (2/2)

zqsz-xiv:
Non-pot potency modifiers are now declared along with skill declarations for all skills, with the exception of T3/high thunder dot nodes. DoTs are rare enough that I think it's fine to force people to explicitly declare all modifiers for those, and would give a relatively small surface to refactor.

Tested visible potency values for some timelines of both PCT and BLM; these changes are purely cosmetic either way.


miyehn (approved):
🚀

PR #48, opened 2024-11-08
Dynamic job swapping

zqsz-xiv:
Adds dynamic job swapping.

image

I've punted on UI decisions in this PR ([see discord discussion](https://discord.com/channels/277897135515762698/1255782490862387360/1297036718972600410)), and just added it as a generic config variable, feel free to push to this branch to change @miyehn (or just change it after this is merged).

Notes:
- A new `job` field is added to the `GameConfig` class, and `ShellVersion` has been incremented.
- Changing the configured job won't take effect until "apply and reset" is hit and the timeline is wiped clean, just as with any other configuration variable.
- There's a minor stutter for icons to reload when a job-swapping configuration is applied--it's not super major since it happens pretty fast, but adding a loading spinner or similar would make it look nicer.
- Any old fight plans in localStorage (or loaded via import) have a job inferred from actions in the timeline.
- Text in `IntroSection.tsx` is updated with an explicit update call from the controller, we may need to double check if this causes any perf/load time regression since it doesn't match up well with the natural react lifecycle. We could also consider just using a single unified `IntroSection` for all jobs (maybe just swapping the help channel/resources links) since we'll all be on the same website.

### DIFF
--- a/src/Controller/Controller.ts
+++ b/src/Controller/Controller.ts
@@ -376,7 +376,6 @@ class Controller {
 		if (content.config.job === undefined) {
 			// infer the job based on actions present
 			content.config.job = inferJobFromActions(content.actions);
-			content.config.shellVersion = ShellVersion.DynamicJob;
 		}
 
 		let gameConfig = new GameConfig(content.config);
--- a/src/Controller/Controller.ts
+++ b/src/Controller/Controller.ts
@@ -6,10 +6,11 @@ import {
 	ShellInfo,
 	ShellJob,
 	ShellVersion,
-	TickMode
+	TickMode,
+	ALL_JOBS,
 } from "./Common";
 import {GameState} from "../Game/GameState";
-import {getAutoReplacedSkillName, getConditionalReplacement, getNormalizedSkillName} from "../Game/Skills";
+import {getAutoReplacedSkillName, getConditionalReplacement, getNormalizedSkillName, jobHasSkill} from "../Game/Skills";
 import {BLMState} from "../Game/Jobs/BLM";
 import {PCTState} from "../Game/Jobs/PCT";
 import {Buff} from "../Game/Buffs";
@@ -20,7 +21,7 @@ import {PCTStatusPropsGenerator} from "../Components/Jobs/PCT";
 import {updateStatusDisplay} from "../Components/StatusDisplay";
 import {updateSkillButtons} from "../Components/Skills";
 import {updateConfigDisplay} from "../Components/PlaybackControl"
-import {setHistorical, setRealTime} from "../Components/Main";
+import {setJob, setHistorical, setRealTime} from "../Components/Main";
 import {ElemType, MAX_TIMELINE_SLOTS, Timeline} from "./Timeline"
 import {scrollTimelineTo, updateTimelineView} from "../Components/Timeline";
 import {ActionNode, ActionType, Line, Record} from "./Record";
@@ -45,10 +46,29 @@ require("../Game/Jobs/RoleActions");
 type Fixme = any;
 
 const newGameState = (config: GameConfig) => {
-	if (ShellInfo.job === ShellJob.PCT) {
+	if (config.job === ShellJob.PCT) {
 		return new PCTState(config);
 	}
 	return new BLMState(config);
+};
+
+const inferJobFromActions = (actions: object[]) => {
+	// Iterate over the whole record and return the number of actions that occur in each job.
+	// If two jobs are tied (like if there's only a Swiftcast/Sprint in the timeline), just
+	// take the first that appears since it doesn't really matter.
+	let maxJob = ALL_JOBS[0];
+	let maxCount = 0;
+	ALL_JOBS.forEach((job) => {
+		const count = actions.reduce(
+			(acc, skill) => (skill as any).skillName && jobHasSkill(job, getNormalizedSkillName((skill as any).skillName)!) ? acc + 1 : acc,
+			0,
+		);
+		if (count > maxCount) {
+			maxJob = job;
+			maxCount = count;
+		}
+	});
+	return maxJob;
 };
 
 class Controller {
@@ -353,6 +373,11 @@ class Controller {
 		if (content.config.shellVersion === undefined) {
 			content.config.shellVersion = ShellVersion.Initial;
 		}
+		if (content.config.job === undefined) {
+			// infer the job based on actions present
+			content.config.job = inferJobFromActions(content.actions);
+			content.config.shellVersion = ShellVersion.DynamicJob;
+		}
 
 		let gameConfig = new GameConfig(content.config);
 
@@ -643,7 +668,7 @@ class Controller {
 			canMove: game.resources.get(ResourceType.Movement).available(1),
 		};
 		if (typeof updateStatusDisplay !== "undefined") {
-			const propsGenerator = ShellInfo.job === ShellJob.PCT
+			const propsGenerator = game.job === ShellJob.PCT
 				? new PCTStatusPropsGenerator(game as PCTState)
 				: new BLMStatusPropsGenerator(game as BLMState);
 			updateStatusDisplay({
@@ -710,6 +735,7 @@ class Controller {
 	}
 
 	setConfigAndRestart(props: {
+		job: ShellJob,
 		level: LevelSync,
 		spellSpeed: number,
 		criticalHit: number,
@@ -724,6 +750,7 @@ class Controller {
 		procMode: ProcMode,
 		initialResourceOverrides: any[],
 	}) {
+		const oldJob = this.gameConfig.job;
 		this.gameConfig = new GameConfig({
 			...props,
 			shellVersion: ShellInfo.version,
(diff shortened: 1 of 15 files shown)


miyehn:
I'm going to advocate for the middle option here [dc message link](https://discord.com/channels/277897135515762698/1255782490862387360/1296834684868759552). I think Turtle's arguments are valid, plus the fact that the information above the current job dropdown also depends on the job dropdown (aka not just numbers, but also what information is shown). <-- I'll push to this branch soon-ish and you can check

And [this](https://discord.com/channels/277897135515762698/1255782490862387360/1296801301954428951) sounds pretty nice too and simple enough to do . <--I'll add afer this is merged

miyehn:
`IntroSection.tsx`: yeah I agree, eventually we can just make it static like before, because it's introducing the all-job rotation planner, not just the currently active job. We don't even need to swap the links and point of contact etc. we can just list them all... there's not going to be that many links I hope?

miyehn:
Anyway what do you think about this:
usually I think this is fine: 
image
then if we change to another job but the change hasn't been applied yet, essentially the summary section is outdated so I strike it out:
image

zqsz-xiv:
I think 🐢's hierarchy argument makes sense. I'd prefer to just not display the irrelevant summary info rather than striking it out so the screen doesn't get too cluttered.

My only concern with this view is that you now need to scroll to see the "apply and reset" button, and the horizontal divider between summary text + other config fields might not make it clear that you need to hit "apply and reset" to change the job. Maybe just adding a text reminder on deteching changed job (like "hit apply and reset below to confirm job change") would work.

src/Controller/Controller.ts
```diff
@@ -353,6 +373,11 @@ class Controller {
 		if (content.config.shellVersion === undefined) {
 			content.config.shellVersion = ShellVersion.Initial;
 		}
+		if (content.config.job === undefined) {
+			// infer the job based on actions present
+			content.config.job = inferJobFromActions(content.actions);
+			content.config.shellVersion = ShellVersion.DynamicJob;
```
miyehn, line 379: We probably shouldn't auto bump `shellVersion` here: unlikely to happen now but if someone loads a very old file with `shellVersion === 0`, it gets bumped to 2 here, then the warning message for converting fixed caster tax to fps tax won't show :0

src/Controller/DamageStatistics.ts
```diff
@@ -167,9 +167,7 @@ function expandNode(node: ActionNode) : ExpandedNode {
 				const tag = modifier.source;
 				if (tag !== PotencyModifierType.ENO && tag !== PotencyModifierType.POT) {
 					res.displayedModifiers.push(tag);
-					if (ShellInfo.job === ShellJob.BLM) {
-						break;
-					}
+					break;
```
miyehn, line 170: we sure other jobs also just always show only the first tag :0? (but also can come back anytime...)

src/Game/GameState.ts
```diff
@@ -49,7 +49,7 @@ export abstract class GameState {
 
 	constructor(config: GameConfig) {
 		this.config = config;
-		this.job = ShellInfo.job; // TODO make this configurable
+		this.job = config.job;
```
miyehn, line 52: I guess... we don't need another state here? because we can just use `config.job`. So there's no risk of the two states that hold the same information get out of sync etc.
(I guess the same applies to `controller.config`... but I've been too lazy to clean stuff up ;_;)

miyehn (approved):
(all are nit picks so)

miyehn:
> I'd prefer to just not display the irrelevant summary info rather than striking it out so the screen doesn't get too cluttered.

I'm fine with both tbh. If we hide the irrelevant summary the only down side is everything will suddenly get moved up and cause a small [layout shift](https://developers.google.com/publisher-tag/guides/minimize-layout-shift), but then yeah less distraction and nobody wonders what strikethrough means

Another thing we can do is to fix the "apply and reset" button to the bottom of that section, and only elements above are scrollable 🤔

src/Controller/DamageStatistics.ts
zqsz-xiv: This part of the loop is guarded by `AFUISkills.has(node.skillName)`, so it's only run for BLM anyway.

src/Game/GameState.ts
zqsz-xiv: Good point, I replaced this with `get job() { return this.config.job; }`.

miyehn:
> and the horizontal divider between summary text + other config fields might not make it clear that you need to hit "apply and reset" to change the job

Another random idea:
image

zqsz-xiv:
> If we hide the irrelevant summary the only down side is everything will suddenly get moved up and cause a small [layout shift](https://developers.google.com/publisher-tag/guides/minimize-layout-shift)

Fair enough. I just realized that the info doesn't update live with config anyway (the cast time table only displays cast times from the confirmed sps). My thought was that if other jobs had additional info they wanted to display in the summary area, we could swap to that as the job dropdown was changed, but that would be inconsistent with other config behavior.

src/Controller/Controller.ts
zqsz-xiv: I think it would be a little weird to have a file with `shellVersion < DynamicJob` that has a `job` field set, since a user then would be able to re-export a timeline made with `shellVersion === 0` but also has a `job` attached. Not sure what the best way to resolve this is--instead of checking if `shellVersion === 0` for printing the warning, can we check the presence of the legacy config field instead?

zqsz-xiv:
> Another thing we can do is to fix the "apply and reset" button to the bottom of that section, and only elements above are scrollable

I think this would make sense, I just had the second realization that expanding the override tab will make the "apply and reset button" go to the bottom as well.

miyehn:
How about... 
* keep the strikethrough (actually I'll modify the logic a bit more)
* change the horizontal bar to the vertical line on the left like in the above screenshot
* fix "apply and reset" at bottom to be always visible

^I'll do these and add the job colors, and check the resource overrides section.... tomorrow, my brain is getting foggy now... 💤

src/Controller/Controller.ts
miyehn: `shellVersion < DynamicJob` but also has `job` attached should be fine I think...? Because all the logic that depends on `shellVersion >= 0` and `shellVersion >= 1` still work. I guess it's fine in general if input data contains something extra: missing data does imply old data version, but extra data doesn't mean anything in particular. But I'm not a webdev and don't work with this kind of data, so I don't actually know the common practice :0

I made `shellVersion` because it lets us make assumptions about data by just checking `shellVersion` alone, instead of having to check potentially multiple fields (`fps` + `gcdCorrection`, having one should imply the other but what if there's a bug and they get out of sync etc), plus more readable. And it's handy when we need to treat old and new data differently, likely because we can't just trivially upgrade old data to the new version and treat them the same

So technically we don't really need to add a new version this time (it doesn't hurt either), because the `job` field can be perfectly inferred from the rest of the file (I think??? Or if it's an empty file it should work for the default BLM job anyway). And once the file is loaded we can just always assume the `job` field exists and read it whenever. Just like when we added `det` and some other stats (just set them to the default and everything still works)

zqsz-xiv: > we don't really need to add a new version this time (it doesn't hurt either), because the job field can be perfectly inferred from the rest of the file

That's true, I'll revert the `ShellVersion` bump in light of this.

My concern is that we have several nullable fields in the config (level, crit, DH, and det) where it's not necessarily clear why they're nullable to begin with: in fact, the type signature of the [constructor](https://github.com/xivintheshell/xivintheshell/blob/47dd02f136e3160cb996ea81e64540700ea08368/src/Game/GameConfig.ts#L86-L100) doesn't even say they can be unset, but since we call the constructor after parsing a raw JSON object [in the controller](https://github.com/xivintheshell/xivintheshell/blob/47dd02f136e3160cb996ea81e64540700ea08368/src/Controller/Controller.ts#L357) they secretly can be undefined. In this particular case it's just some type system escsape hatches being used that we should probably rewrite, but in general I'm worried that if we have serialization changes independent of the `shellVersion` (or if a field needs to be parsed differently across different versions), it becomes hard to track what changes need to be made. That said, for these fields it's obvious enough what needs to change, and like you said we can always infer the job of an older file.

I'm not sure what the "usual" way of modifying frontend-only persisted state is, but in my experience with versioned data on the backend, you usually define a sequence of "migrations" that describe the changes that need to be made between versions. For instance, if you have a struct `{ version: 0, a: string }` that you add a field to (becoming `{ version: 1, a: string, b: number }`), you would have a migration function that populates `b` if it detects `version == 0`. If you then make a later change to remove a field (like `{ version: 2, b: number }`), you would then have a second migration that takes in an object with `version == 1` and removes `a` from the data. If you need to migrate an object from `version == 0` to `version == 2` it would just apply each migration in sequence.

The code we have in the controller + config kind of does this, just checking for unset fields rather than checking a version. Personally I'd be more comfortable if all changes to the config were assigned a version so we could have very explicit migrations and keep it clear what fields we would expect a version of config to have (useful in case if we added a bad config field that we need to revert), but the config should change infrequently enough that what we have is fine and we don't need to overengineer it.

miyehn: If the types don't make sense somewhere you're probably right... I just looked at the GameConfig constructor and saw parts of it last modified more than two years ago... probably either something I migrated from js to ts without giving too much thought, or I didn't know better (this was my first time using typescript) and I'm always too lazy to revist (monkeylookingaway.gif)
For file versioning I agree it'll be much more robust written in the way you described. Currently it's assuming extra data doesn't do any harm and most version changes can just upgrade silently (which has been the case). Right now it doesn't cause any trouble so I'm definitely too lazy to rewrite, but if you have the time and energy to spare then by all means xD

zqsz-xiv: understandable

> if you have the time and energy to spare then by all means

Maybe later but let's just leave this be for now :P

miyehn:
fixing the button position wasn't as trivial as I thought so didn't do it just now.. anyway I think I'll leave the config tab and go mess with colors in another branch

zqsz-xiv:
sgtm, feel free to hit the merge button yourself if you're happy with the frontend for now. Thanks as always for reviewing + editing.

Are we planning on keeping this out of deployment until we're running on the new domain?

miyehn:
We can deploy it now and start testing things out. But probably shouldn't announce publicly yet, not until we have the localStorage migration part figured out

PR #49, opened 2024-11-09
clean up intro section and make favicon change

zqsz-xiv:
Acknowledgements section is now unified (we can consider updating contacts later if we end up with more maintainers for other jobs).

Removed the cross-links for BLM/PCT in the Shell + Akairyu's lvl 90 fork, now it's just github repo + EW archive + bozjan shell + Balance job resources channel.

The favicon now also changes when the job is changed.


miyehn:
I propose that we unify them completely so acknowledgements section doesn't depend on job at all, and also change titles "* in the Shell" to just "FFXIV in the Shell". Because currently applying a new config updates the section and favicon etc. but switching to a slot with a different job doesn't and it feels a bit strange 🤔

zqsz-xiv:
> switching to a slot with a different job doesn't

Oops, that's an oversight on my part. I can fix that later.

> I propose that we unify them completely so acknowledgements section doesn't depend on job at all

What do you think we should do about job-specific content (pretty much just the Balance job resources channel + official job guide)?

miyehn:
> the Balance job resources channel + official job guide

We can get rid of them if there are already too many links, or just link to the server + [page with links to all job guides](https://na.finalfantasyxiv.com/jobguide/battle/)

miyehn:
> Oops, that's an oversight on my part. I can fix that later.

I think switching slot not updating job is fixed as a side effect of my last commit.. 😅  (those global frontend upate functions would also benefit a lot from a cleanup pass but meh...)

zqsz-xiv:
Seems like it did get fixed.

I added links to the job guide homepage like you selected, plus links to the Balance + Balance discord (pending localization). I changed the h1 to just be "FFXIV in the Shell", and also updated the page's `<title>` to include the job in parentheses. I left in the "implementation notes" section for BLM, not sure what you want to do with that.

At some point we might want to do a revamp of the instructions + about sections to be a pop-up overlay instead of an expandable element: that way we wouldn't have to worry about changing it in sync with the active timeline either.

miyehn (approved):
> implementation notes for BLM

I'm not sure what I want to do with that either.... guess I'll leave it be for now

PR #50, opened 2024-11-09
Active timeline slot handle color based on job; PCT resource color adjustments

miyehn:
Timeline slot handle:
image
image
I kept inactive slot grey because job colors at low transparency either look very dull and hard to differentiate anyway, or too bright and hard to tell if the slot is active or not. 

I'm also proposing some resource color adjustments for PCT referencing the job gauge, mostly checked their contrast against the background to make them readable & recognizable:
image

(now that I look at it again I think holy could be more blue and slightly darker... but lmk if the general direction looks okay, especially if people are used to the current colors already)

### DIFF
--- a/src/Components/ColorTheme.tsx
+++ b/src/Components/ColorTheme.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import {MdDarkMode, MdLightMode} from "react-icons/md";
-import {getCachedValue, setCachedValue} from "../Controller/Common";
+import {getCachedValue, setCachedValue, ShellJob} from "../Controller/Common";
 import {controller} from "../Controller/Controller";
 
 export type ColorTheme = "Light" | "Dark";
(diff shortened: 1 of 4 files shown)


zqsz-xiv:
Resource colors lgtm.

Where else is the accent color used? Is it just buff highlights + checkbox/radio box fills?

zqsz-xiv (approved):

miyehn:
> Where else is the accent color used?

Border for skills input region. 

The concern is for some jobs the accent color might collide (or not have enough contrast) with other color-coded information, for example

links
image

Warning, historical, pot
image
image

zqsz-xiv:
I think keeping the primary color as purple for all jobs is fine, tbh I didn't even realize most of these were meant to be job-specific. Splitting colors by job might also make future additions to the timeline harder: besides the examples you gave, I was thinking that if we ever want to support partially invalid timelines, we might want to highlight invalid skill usages in red (which would be confusing if we had a job whose accent color is red).

PR #53, opened 2024-11-10
Update most icons to pull from xivapi instead of local assets

AkairyuGestalter:
The deeper we get into having multiple jobs under the xivintheshell banner, the more bloated the repository (and the deployed site) is going to get if we keep exporting assets into source. I pulled the icon URL calculation code from xivanalysis and made some updates to allow us to use xivapi as the icon source for most icons, rather than local assets.

I didn't deal with the buff icons that show up on the action tooltip in the timeline and whatnot yet, this was mainly to update the action and status buff icons for the active timeline row's action selection and status effect display. I also didn't remove the ability to have local assets, since we might still need that in the future for other purposes or for things like the "missing skill" icon which I didn't feel like trying to hunt down a xivapi icon number for.


AkairyuGestalter:
We can resurrect this later if there's appetite for it, but for now, abandoning the idea because of potential CN firewall nonsense I'd forgotten we discussed previously on Discord.

miyehn:
Link to discussion on Discord for future reference: https://discord.com/channels/277897135515762698/1255782490862387360/1305270539949969419

PR #54, opened 2024-11-11
RDM in the Shell

zqsz-xiv:
Testing adding a new job under the current framework.

Everything is complete except for better theme colors, which I'll get to tomorrow.

My thoughts on ease of implementation (obviously biased because I'm very familiar with the codebase): 95% of this was pretty easy to implement, as the new skill/resource interfaces are very declarative and easy to use. The last 5% was a little annoying and generally involved a lot of copy-paste and busywork.
- Downloading/naming icons is annoying, though the new script I added should fix this.
- Making a buff show up as a "cover" in the timeline + hover on mouseover requires digging into the weeds a bit.

Other things added in this PR:
- Combos can just be built on top of the existing resource system with some modifiers to helper functions; also introduces "additive" potency modifiers (thank you to Sterling Sylva @ Goblin for implementing this in SAM in the Shell)
- Splitting up the `SkillName` and `ResourceType` into separate files for each job. Everything can still be accessed under the original enum namespaces, though some type annotations may need to be changed.


miyehn (approved):
I have pretty much 0 knowledge about RDM but it's not breaking BLM (further) and I trust your judgement 🙏 
This can go in. I'll fill in translation when I get to it or I summon Galahad

I did find in this branch that BLM files with old ShellVersion loads with a bunch of console errors. It's not introduced by this PR though.. we fix it when we get to it:
[7.05绝欧P6延后魔泉轴2.44.txt](https://github.com/user-attachments/files/17697480/7.05.P6.2.44.txt)
* To be specific, happens when we load this file into a slot that is currently set to another job

zqsz-xiv:
Sounds good, I'll merge this for now then. I covered a pretty solid amount of edge cases, and if anything is broken I can take care of it later.

I also found a bug where you get an exception (in production it's just a console error) when you try to load a preset line with skills for a different job, I'll fix that in a separate PR.

PR #65, opened 2024-11-17
DNC in the shell

AkairyuGestalter:
You can shell DNC if you want to, you can leave spreadsheets behind~

I updated the URL that the icon downloader script pulls from to use the updated xivapi beta site, which is the one that's actually being maintained going forward.

Also included is some work to allow actions to affect/be affected by multiple cooldowns for things like GCDs with an associated longer cooldown (ie. Standard/Technical/MCH tools), or abilities that share a cooldown in-game (ie. Closed Position and Ending).

Longer term, we should probably refactor this to more closely match game state which defines cooldowns by group, and any given action can be affected by an array of coldowngroups, but this should be sufficient for any near-term use.


miyehn:
:0 just poked around a bit and I saw this:
image
I don't think we should use the green GCD bar to indicate anything other than GCD. The green GCD bars and purple cast bars are displayed like they do now to indicate whichever is longer will block the next skill/GCD, but in this case this green bar isn't blocking anything. And may overlap with later GCDs and make it hard to tell how long a skill takes.

I don't play dancer and maybe it doesn't matter as much specifically for dancers, but I think we should keep the visuals consistent whenever possible, so players using other jobs don't get confused.
* How is ths status shown in game? Is it a buff? If so can we also switch to using a buff instead and get rid of this buff after `Standard Finish`?

AkairyuGestalter:
That's probably an unintentional side effect of setting up the additional cooldown handling for Standard's 30s cooldown... I didn't really pay attention to that since I was focused on making things work right. I'll take a look to see if I can figure out how to fix that

zqsz-xiv (commented):
Got a few comments, otherwise LGTM! Can you try to also fix the indentation in `DNC.ts`? It seems like there's a mix of spaces/tabs that makes it hard to read in the github editor (partly my fault, since my editor is set to spaces by default and I think miyehn uses tabs, and I forgot to convert indentation on some of the new files I made).

src/Game/Jobs/DNC.ts
```diff
@@ -0,0 +1,985 @@
...
+    potency: 350,
+    applicationDelay: 0.54,
+    onConfirm: (state) => {
+        state.tryConsumeResource(ResourceType.TechnicalStep)
+        state.tryConsumeResource(ResourceType.TechnicalDance, true)
+    },
+    recastTime: 1.5,
+    validateAttempt: (state) => state.hasResourceAvailable(ResourceType.TechnicalStep),
+})
+makeGCD_DNC(SkillName.SingleTechnicalFinish, 70, {
```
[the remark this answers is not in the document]
AkairyuGestalter: I kept Quadruple Technical Finish split out cause it has some extra handling, but not a bad idea for the others. Done.

src/Game/Jobs/DNC.ts
```diff
@@ -0,0 +1,985 @@
...
+
+        let rand = this.rng();
+        if (this.config.procMode === ProcMode.Always || rand < chance) {
+            this.gainProc(proc)
+        }
+    }
+
+    gainResource(rscType: typeof ResourceType.EspritGauge | typeof ResourceType.FeatherGauge, amount: number) {
+        const resource = this.resources.get(rscType)
+        if (resource.availableAmount() === resource.maxValue) {
```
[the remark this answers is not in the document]
AkairyuGestalter: Yeah, it should be (more that you can be at 95 Esprit and gain 10). I did Feathers first which this works fine for, but Esprit breaks that original coding.

src/Game/Jobs/DNC.ts
```diff
@@ -0,0 +1,982 @@
...
+    secondaryCooldown: {
+        cdName: ResourceType.cd_TechnicalStep,
+        cooldown: 120,
+        maxCharges: 1,
+    },
+})
+makeGCD_DNC(SkillName.Tillana, 82, {
+    startOnHotbar: false,
+    potency: 600,
+    recastTime: (state) => state.config.adjustedSksGCD(),
```
zqsz-xiv, line 683: Doesn't Tillana not scale to your GCD in 7.x?

src/Components/PlaybackControl.tsx
```diff
@@ -493,6 +500,28 @@ export class Config extends React.Component {
 			});
 		}
 
+		this.updateSksTaxPreview = (sksStr: string, fpsStr: string, levelStr: string) => {
```
zqsz-xiv, line 503: Looks like you missed a call to this in `this.setJob`, since the preview's showing as "n/a" on page reload.

zqsz-xiv (commented):
Poked around a bit more, found a couple console errors:
- Need to register `TechnicalBonus` and `StandardBonus` ResourceTypes
- Need to stop the lucid tick event from being registered in `GameState.ts` (see https://github.com/xivintheshell/xivintheshell/blob/fc02ef11fb42d2ed43bc55bb5c809c649c274dfc/src/Game/GameState.ts#L183-L188), I'll refactor it out later
I'm also getting a `Cannot read properties of undefined (reading 'warningOnTimeout')` from [this line](https://github.com/xivintheshell/xivintheshell/blob/3f5da45ab177f676a8b46d722eade5b03496f783/src/Game/GameState.ts#L590) in `GameState::enqueueResourceDrop`. I think it's caused by those resources not being registered, but if you could add a `?` to `rscInfo.warningOnTimeout` to prevent the exception that would be helpful (missing resources already raise a console error).

Also, please double-check indentation (or just get your editor to do a conversion) in `Game/Jobs/DNC.ts`, some stuff is showing up weirdly in the github editor and it looks like there's still a mix of tabs/spaces
image.

zqsz-xiv:
Here's the file I'm using to get the exception, I see the exception after loading this and then pressing any GCD.

[dnc_error_after_one_more_button.txt](https://github.com/user-attachments/files/17808920/dnc_error_after_one_more_button.txt)

src/changelog.json
```diff
@@ -1,4 +1,10 @@
 [
+	{
+		"date": "11/17/24",
```
zqsz-xiv, line 3: remember to update this with whatever date it gets merged, right now react is complaining because it's duplicating a key from a different changelog entry

src/Components/PlaybackControl.tsx
```diff
@@ -548,18 +577,23 @@ export class Config extends React.Component {
 							throw new Error("Imported gearset was for a job (" + body["jobAbbrev"] + ") that XIV in the Shell doesn't support");
 						}
 						// TODO should probably validate each of these fields
+						const baseSpeed = XIVMath.getSubstatBase(parseFloat(body["level"]))
+						const spellSpeed = stats.get("SPS") ?? baseSpeed
```
miyehn, line 581: nit: type looks kinda... unnecessarily mixed in these couple of lines. Maybe just stick to either string or number? And would also be able to get rid of the exclamation marks on lines 595 and 596 🤔

src/Components/PlaybackControl.tsx
```diff
@@ -548,18 +577,23 @@ export class Config extends React.Component {
...
 							level: body["level"],
-							spellSpeed: stats.get("SPS"),
+							spellSpeed,
+							skillSpeed,
 							criticalHit: stats.get("CRT"),
 							directHit: stats.get("DH"),
 							determination: stats.get("DET"),
 							imported: true,
-							importedFields: ["job", "level", "spellSpeed", "criticalHit", "directHit", "determination"],
+							importedFields: ["job", "level", "spellSpeed", "skillSpeed", "criticalHit", "directHit", "determination"],
```
miyehn, line 592: Also, does etro not always import both SPS and SKS? Does that mean only one of the two should be in `importedFields` here? :0

miyehn (changes requested):
Not done reviewing yet but:
image
I guess DNC shouldn't be ticking lucid dreaming? 😂

miyehn:
Also no.... let's not leave the huge `minHeight` as is... I will figure something out......

src/Components/PlaybackControl.tsx
zqsz-xiv: It looks like you're right, etro gives only one of the two, but xivgear gives both.

miyehn:
(continuation from discord)
Seems like even to do something special for DNC is going to require a tiny bit of refactor... I guess, for this PR can we have the huge `minHight` just for DNC but not other jobs? And I'll make another PR to implement this:
image
Or alternatively merge the `minHeight` as-is but don't deploy yet...

src/Game/GameConfig.ts
```diff
@@ -131,15 +137,19 @@ export class GameConfig {
...
 	}
 
-	adjustedDoTPotency(inPotency : number) {
-		return XIVMath.dotPotency(this.level, this.spellSpeed, inPotency);
+	adjustedDoTPotency(inPotency : number, affectedBySkillSpeed: boolean = false) {
+		return XIVMath.dotPotency(this.level, affectedBySkillSpeed ? this.skillSpeed: this.spellSpeed, inPotency);
 	}
 
 	// returns GCD before FPS tax
 	adjustedGCD(baseGCD: number = 2.5, speedBuff?: ResourceType) {
```
miyehn, line 145: nit: we can rename this one too, `adjustedGCD` -> `adjustedSpsGCD`

miyehn (commented):
I didn't check the job implementation specifics but others changes mostly look good to me👌

src/Components/PlaybackControl.tsx
```diff
@@ -466,6 +471,8 @@ export class Config extends React.Component {
 			b1TaxPreview: "n/a",
 			gcdPreview: "n/a",
 			taxedGcdPreview: "n/a",
+			sksGcdPreview: "n/a",
```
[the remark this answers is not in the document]
AkairyuGestalter: I was somewhat intentionally leaving the existing properties as-is so we didn't have to do change them everywhere else.

src/Components/PlaybackControl.tsx
AkairyuGestalter: the exclamation points were already there before I make the change, and I'm not familiar with what that notation doeas, so I'll leave that alone, but I can at least toString the retrieval of the base speed so we're always dealing with strings for the speed stats themselves

src/changelog.json
AkairyuGestalter: Done

src/Components/PlaybackControl.tsx
AkairyuGestalter: Yeah, I'd missed that when I was creating the sks version. Fixed.

src/Game/Jobs/DNC.ts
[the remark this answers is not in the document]
AkairyuGestalter: Actually ended up compressing both Standard and Technical finish action creation into their own array foreaches, with some conditionals to handle bailing early on no-steps and figuring out which bonus level to apply dynamically based on executed steps

AkairyuGestalter:
Changes should all be present. If there are more indent level problems we should consider setting up eslint rules to enforce whichever flavor we want, and can also set up vscode rules to automatically fix lint errors on save so we don't have this problem in the future.

The minHeight change is being left in for now since the code locus that was made at has no notion of the current job, so I can't condition it as such. Plus it seems there's other ideas baking anyways...

src/Components/PlaybackControl.tsx
zqsz-xiv: `!` is a type system escape hatch that asserts the value cannot be null/undefined. It's no longer necessary here because `stats.get("SPS") ?? baseSpeed` cannot be undefined, where previously `stats.get("SPS")` could have returned undefined.

zqsz-xiv:
> If there are more indent level problems we should consider setting up eslint rules to enforce whichever flavor we want, and can also set up vscode rules to automatically fix lint errors on save so we don't have this problem in the future.

Yeah... we should probably wait until a couple weeks after FRU and the codebase settles down so we don't introduce a bunch of merge conflicts due to linting though. Might also want to make precommit hook for this purpose.

> The minHeight change is being left in for now since the code locus that was made at has no notion of the current job,

We have a hacky way to get the currently displayed job with `controller.getActiveJob()`. It's unidiomatic but it should work, and I think it's worth doing because the big minheight does make the other jobs look kinda weird.

zqsz-xiv:
Also, damage table still isn't rendering modifiers correctly: should be an easy fix, you'll just need to modify the `buffName` and `BuffTag` functions in `src/Components/DamageStatistics.tsx` (https://github.com/xivintheshell/xivintheshell/wiki/Adding-a-new-job#the-damage-table).

image

src/Components/DamageStatistics.tsx
```diff
@@ -544,10 +579,10 @@ export class DamageStatistics extends React.Component {
 			return <div key={props.key} style={rowStyle}>
 				<div style={cell(3)}>{includeCheckboxes}</div>
 				<div style={cell(18)}>{skillNameNode}</div>
-				<div style={cell(14)}>{tags}</div>
+				<div style={cell(19)}>{tags}</div>
```
miyehn, line 582: I will get home later and check if changing the table layout is okay here- the previous ones worked fine for BLM, PCT and RDM, maybe DNC just has more buffs.
But also, if possible I'd advocate for having as few tags as possible, and only leave the ones that are actually helpful. There was a related discussion [here ](https://github.com/xivintheshell/xivintheshell/pull/43#pullrequestreview-2383992217)(need to expand resolved)

src/Game/Potency.ts
```diff
@@ -112,8 +112,8 @@ export const Modifiers = {
 	Devilment: {
 		kind: "critDirect",
 		source: PotencyModifierType.DEVILMENT,
-		critFactor: 1.2,
-		dhFactor: 1.2,
+		critFactor: 0.2,
```
miyehn, line 115: rip time for us to rename them to be clear :0

miyehn (approved):
Anyway I might clean up some nits myself later but I don't see more blocking issues 🙏

miyehn:
image
There's this but should be quick to fix; I can fix it later too

PR #67, opened 2024-11-21
Fix Standard/Technical Finish highlighting

AkairyuGestalter:
The finishes were validating on having the right number of dance moves, instead of highlighting on it. Technically it's the result you want, but it's not accurate to game behavior!


miyehn (approved):
🚀

PR #69, opened 2024-11-23
Refactor TraitList into Jobs/Common.ts

misterperson:
Not sure if there's a better way to do this that avoids an apparent circular import between Common and Traits.
I tried to define TraitName elsewhere and only build the map from Trait.ts, but I was running into static initialization errors with `Trait:buildList()` referencing `Common.ts:TraitList` before it was initialized.

### DIFF
--- a/src/traits.test.js
+++ b/src/traits.test.js
@@ -1,0 +1,53 @@
+import {TraitName,ReservedTraitName} from "./Game/Common"
+import {Traits} from "./Game/Traits"
+
+function GetTraits() {
+    const traitNameArr = Object.keys(TraitName).filter((v) => isNaN(Number(v)))
+    const traitNumberArr = Object.values(TraitName).filter((v) => !isNaN(Number(v)));
+    return  traitNameArr
+        .map((e, i) => [e, traitNumberArr[i]])
+        .filter(([name, value]) => !(Object.keys(ReservedTraitName).includes(name)));
+}
+
+// for helpful error messages
+expect.extend({
+    toBeGreaterExt([name, value], expected) {
+        return {
+            message: () => `${name} = ${value}. Must be greater than ${expected} (values 0-${expected} are reserved)`,
+            pass: value > expected
+        }
+    },
+    toBeUnique([name, value], map) {
+        return {
+            message: () => `${name} conflicts with ${map.get(value)}. Both have a value of ${value}.`,
+            pass: !map.has(value)
+        }
+    },
+    toBeLockedAtLevel0([name, value]) {
+        return {
+            message: () => `${name} is missing from Traits lookup. Make sure it is included in the TraitList`,
+            pass: !Traits.hasUnlocked(value, 0)
+        }
+    }
+  });
+
+// If this test fails, a Job's TraitName enum start was not set
+it("has no traits with value less than 100", async () => {
+    GetTraits().forEach(([name, value]) => {
+        expect([name, value]).toBeGreaterExt(99);
+    })
+})
+
+it("has unique trait values", async () => {
+    let map = new Map();
+    GetTraits().forEach(([name, value]) => {
+        expect([name, value]).toBeUnique(map);
+        map.set(value, name);
+    })
+})
+
+it("allows lookup of all traits", async () => {
+    GetTraits().forEach(([name, value]) => {
+        expect([name,value]).toBeLockedAtLevel0();
+    });
+})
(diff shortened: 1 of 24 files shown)


zqsz-xiv (changes requested):
Thanks for the PR!

> I was running into static initialization errors with Trait:buildList() referencing Common.ts:TraitList before it was initialized

Would it be possible to split `TraitName` into stuff like `BLMTraitName` up like we do for `SkillName`, and keep declarations like `BLMTraitList` as before? You should be able to refer to `BLMTraitName` as the keys of the `TraitList` objects, possibly with some extra casting in the union in `Common.ts`.

misterperson:
> Would it be possible to split `TraitName` into stuff like `BLMTraitName` up like we do for `SkillName`, and keep declarations like `BLMTraitList` as before? You should be able to refer to `BLMTraitName` as the keys of the `TraitList` objects, possibly with some extra casting in the union in `Common.ts`.

Maybe. My main reason for not splitting up` TraitName` was that numeric enums would start at 1 and count up, so joining multiple enums would result in duplicate keys in the global TraitList. Although, that does point out the dubious nature of a global trait list. Tank Melee Mastery (Endwalker) is unlocked prior to Melee DPS Melee Mastery (Dawntrail).

I'll try splitting it up in a separate commit and get feedback.

misterperson:
I do have a good question about shared trait names...

Tanks get Melee Mastery at level 84
Dragoon gets Melee Mastery at level 94
Monk gets Melee Mastery at 84

Healers currently have a Magick Mastery at 94
Wouldn't put it past SE to add a Magick Mastery to a magical ranged job later on...

How should we differentiate these traits? I know that for EW and DT, the potency trait for each job is at level X4. But traits below 80 are just plain inconsistent (although they are at least unique AFAIK).

zqsz-xiv:
> My main reason for not splitting up TraitName was that numeric enums would start at 1 and count up, so joining multiple enums would result in duplicate keys in the global TraitList

Fair point, I forgot that since traits aren't exposed to users we don't make it a string enum. One possible workaround would be to start the discriminant at a really high value, like giving BLM `EnhancedEnochianII = 0`, PCT `PictomanyMasteryII = 100`, RDM `EnhancedDisplacement = 200`, etc. so they never conflict assuming there won't be 100 traits per job, but this is a bit hacky.

> I'd personally lean towards something like MeleeMasterMNK/MeleeMasterDRG and the like where jobs have the same trait name at different levels, or MeleeMasteryTank if there's a shared name where the whole role gets it at once.

Agreed, this seems reasonable. We should also add a role trait section for stuff like Enhanced Addle/Feint.

misterperson:
Ugh, random whitespace changes. Does this repository have a formatter?

misterperson:
This needs a test to verify that all enum members have unique names, but I'm not sure it would quite work the way that's expected in Typescript since enums are really just Objects.

src/traits.test.js
```diff
@@ -0,0 +1,53 @@
...
+    let map = new Map();
+    GetTraits().forEach(([name, value]) => {
+        expect([name, value]).toBeUnique(map);
+        map.set(value, name);
+    })
+})
+
+it("allows lookup of all traits", async () => {
+    GetTraits().forEach(([name, value]) => {
+        expect([name,value]).toBeUnlockedAtLevelCap();
```
misterperson, line 51: Oh, this test doesn't work as expected because `hasUnlocked` returns `true` for non-existent traits.

zqsz-xiv:
> Does this repository have a formatter?

Unfortunately no, we didn't have one prior to the BLM/PCT merge, and we decided to hold off on adding one since a couple other people were working on additional jobs and we didn't want to create merge conflicts for them. I'll look into setting it up some time after FRU release when things have calmed down a bit.

misterperson:
Updated for RPR

miyehn:
@zqsz-xiv (psst, in case you haven't seen the updates yet 👀)

zqsz-xiv:
Sorry, got waylaid by FRU prog. Overall looks good to me, just two comments:
- Is there any particular reason for starting BLM traits at 9001? The prefixes aren't contiguous anymore and it might cause a bit of confusion when we add our next couple jobs.
- I'd prefer for trait numbers to start 0-indexed (e.g. Never = 0, then each job starts at 1000/2000, etc. rather than 1001/2001).

misterperson:
BLM can be reordered to 1000-1999. Traits can be changed to zero-indexed. Just wanted to insert some humor into the code :p (see `./Constants/BLM.ts`)

The unit tests for traits will verify there are no conflicts between the indices of traits when additional jobs are added.

I can get a fix up pushed later today. Currently stuck at the office writing performance reviews and waiting for a car to charge.

________________________________

zqsz-xiv (approved):
Looks good, thanks!

PR #70, opened 2024-11-23
MCH in the shell

AkairyuGestalter:
In addition to the raw MCH logic, a few things done to core behavior:
- Moved all makeResource calls for role action related buffs into RoleActions.ts alongside their makeResourceAbility calls
- Tweaked DoT display logic to make it a bit easier to maintain the list of actions that qualify
- Edited the potency display table to allow for hiding the "per hit" potency of Wildfire and Queen's actions, since they're variable
- Added a potency modifier type for skills that cannot crit or direct hit, as is the case for Wildfire's on-expiration damage


src/Game/Jobs/MCH.ts
```diff
@@ -0,0 +1,856 @@
...
+        if (!Traits.hasUnlocked(TraitName.QueensGambit, config.level)) {
+            this.resources.set(new Resource(ResourceType.QueenFinishers, 1, 0))
+            this.resources.set(new Resource(ResourceType.Queen, 6, 0))
+        }
+
+        if (!Traits.hasUnlocked(TraitName.EnhancedReassemble, config.level)) {
+            this.cooldowns.set(new CoolDown(ResourceType.cd_Reassemble, 55, 1, 1))
+        }
+        if (!Traits.hasUnlocked(TraitName.EnhancedMultiWeapon, config.level)) {
+            this.cooldowns.set(new CoolDown(ResourceType.cd_Drill, 20, 1, 1))
```
zqsz-xiv, line 78: isn't the recast scaled to sks before you get 2 charges?

src/Game/Jobs/MCH.ts
```diff
@@ -0,0 +1,856 @@
...
+    potency: [
+        [TraitName.Never, 220],
+        [TraitName.MarksmansMasteryII, 240]
+    ],
+    applicationDelay: 0.85,
+    recastTime: 1.5,
+    onConfirm: (state) =>  {
+        (state.cooldowns.get(ResourceType.cd_DoubleCheck) as CoolDown).restore(state, 15);
+        (state.cooldowns.get(ResourceType.cd_Checkmate) as CoolDown).restore(state, 15);
+        state.tryConsumeResource(ResourceType.Overheated)
```
zqsz-xiv, line 570: can other skills consume Overheated if you use them with the buff active?

src/Components/DamageStatistics.tsx
```diff
@@ -506,7 +507,16 @@ export class DamageStatistics extends React.Component {
...
 
 		let isDoTProp = function(skillName: SkillName) {
-			return skillName === SkillName.Thunder3 || skillName === SkillName.HighThunder || skillName === SkillName.Higanbana;
+			return DOT_SKILLS.includes(skillName)
+		}
+
+		let hidePotency = function(skillName: SkillName) {
+			if (isDoTProp(skillName)) { return true}
+			const hidePotencySkills: SkillName[] = [
+				// MCH: Queen and Wildfire have variable potencies per "tic" so just don't show them in the per-hit potency column
```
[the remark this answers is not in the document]
zqsz-xiv: would it be possible to mention this as a help tooltip in the potency column for these 2 skills?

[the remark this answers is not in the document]
zqsz-xiv: also, not quite sure how you're implementing the tick tracking, but the entire queen usage seems to be considered "untargetable" and annulled if one of its hits happens during untargetability (I didn't check wildfire detonation, but since the potency of that isn't applied until the debuff expires then this behavior makes sense for it)

image

src/Components/Jobs/MCH.tsx
```diff
@@ -0,0 +1,207 @@
...
+        {
+            kind: "bar",
+            name: localize({
+                en: "Battery",
+            }),
+            color: colors.mch.battery,
+            progress: battery / 100,
+            valueString: battery.toFixed(0),
+        } as ResourceBarProps,
+        {
```
zqsz-xiv, line 195: would it be better to display the timer bar instead to better match what's shown in-game?

src/Controller/Controller.ts
zqsz-xiv, line 1: I think your editor converted indents to spaces, could you revert it back to tabs to keep the diff small?

src/Game/Constants/MCH.ts
```diff
@@ -0,0 +1,84 @@
...
+    cd_Chainsaw = "cd_Chainsaw",
+    cd_BarrelStabilizer = "cd_BarrelStabilizer",
+    cd_Wildfire = "cd_Wildfire",
+    cd_Queen = "cd_Queen",
+    cd_Overdrive = "cd_Overdrive",
+    cd_Dismantle = "cd_Dismantle",
+    cd_Tactician = "cd_Tactician",
+    cd_Hypercharge = "cd_Hypercharge",
+    cd_Detonator = "cd_Detonator",
+    cd_Robot = "cd_Robot"
```
zqsz-xiv, line 83: is this used anywhere?

src/Game/Jobs/MCH.ts
```diff
@@ -0,0 +1,856 @@
...
+            }));
+        };
+
+        let timeTillFirstDotTick = this.config.timeTillFirstManaTick + this.dotTickOffset;
+        while (timeTillFirstDotTick > 3) timeTillFirstDotTick -= 3;
+        this.addEvent(new Event("initial DoT tick", timeTillFirstDotTick, recurringDotTick));
+    }
+
+    processComboStatus(skill: SkillName) {
+        if (!COMBO_GCDS.includes(skill)) { return; } // MCH's non-combo GCDs don't break an ongoing combo
```
zqsz-xiv, line 118: does pressing the AoE filler break combo?

src/Game/Jobs/MCH.ts
```diff
@@ -0,0 +1,856 @@
...
+        const duration = (getResourceInfo(ShellJob.MCH, proc) as ResourceInfo).maxTimeout;
+        if (this.resources.get(proc).available(1)) {
+            this.resources.get(proc).overrideTimer(this, duration);
+        } else {
+            this.resources.get(proc).gain(1);
+            this.enqueueResourceDrop(proc, duration);
+        }
+    }
+
+    handleQueenPunch = () => {
```
zqsz-xiv, line 150: any particular reason this is an arrow function?

src/Game/Jobs/MCH.ts
```diff
@@ -0,0 +1,856 @@
...
+
+makeAbility_MCH(SkillName.GaussRound, 15, ResourceType.cd_DoubleCheck, {
+    autoUpgrade: {
+        trait: TraitName.DoubleBarrelMastery,
+        otherSkill: SkillName.DoubleCheck
+    },
+    potency: 130,
+    applicationDelay: 0.71,
+    cooldown: 30,
+    maxCharges: 2, // TODO
```
zqsz-xiv, line 584: is the TODO still needed?

src/Game/Jobs/MCH.ts
```diff
@@ -0,0 +1,856 @@
...
+        },
+        onConfirm: (state, node) => {
+            // Cache the battery bonus scalar based on the amount of battery gauge available
+            const battery = state.resources.get(ResourceType.BatteryGauge).availableAmount()
+            state.resources.get(ResourceType.BatteryBonus).gain(battery - 50)
+
+            // Consume the gauge
+            state.tryConsumeResource(ResourceType.BatteryGauge, true)
+
+            // note that queen is summoned, and qrant the requisite number of punches and finishers
```
zqsz-xiv, line 651:
```suggestion
            // note that queen is summoned, and grant the requisite number of punches and finishers
```

src/Game/Jobs/MCH.ts
```diff
@@ -0,0 +1,856 @@
...
+    applicationDelay: 0.89,
+    recastTime: 1.5,
+    onConfirm: (state) =>  {
+        state.tryConsumeResource(ResourceType.Overheated)
+    },
+    validateAttempt: (state) => state.hasResourceAvailable(ResourceType.Overheated),
+    highlightIf: (state) => state.hasResourceAvailable(ResourceType.Overheated),
+})
+
+// Flamethrower - Not adding unless a notable use-case for it is found. right now it's just end-of-downtime tic fishing
```
zqsz-xiv, line 843: if you want to add flamethrower, you should be able to copy most of the SAM meditate code for tick tracking, though this would be a bit more complicated since you would have to register a DoT effect

src/Game/XIVMath.ts
```diff
@@ -32,8 +32,15 @@ export class XIVMath {
...
 		let modifier = damageFactor;
 
-		const critRate = (critBonus >= 1) ? critBonus : XIVMath.#criticalHitRate(level, crit) + critBonus;
-		const dhRate = (critBonus >= 1) ? dhBonus : XIVMath.#directHitRate(level, dh) + dhBonus;
+		const critRate = (critBonus >= 1) ? critBonus : 
+			critBonus < 0 ? 0 : 
+			XIVMath.#criticalHitRate(level, crit) + critBonus;
+		const dhRate = (critBonus >= 1) ? dhBonus : 
+			dhBonus < 0 ? 0 :
+			XIVMath.#directHitRate(level, dh) + dhBonus;
```
zqsz-xiv, line 40: can ou add some extra parentheses here? the associativity of ternary operators is hard to tell at a glance

zqsz-xiv (changes requested):
Thanks! Left some job-specific questions, and once again some editor formatting shenanigans.

Did you test the snapshotting behavior of party buffs for queen + WF? It's not the end of the world if they don't work (since they won't affect the ability to construct timelines), but if they don't line up with in-game behavior we should track it to fix later.

src/Game/Jobs/MCH.ts
AkairyuGestalter: It sure is, good catch.

src/Game/Jobs/MCH.ts
AkairyuGestalter: They do yeah, forgot to account for them since I did them towards the end.

src/Game/Jobs/MCH.ts
AkairyuGestalter: Other places where we were having an event callback recursively call itself used that pattern, like with the DoT tick handlers.

src/Game/Jobs/MCH.ts
AkairyuGestalter: Yep, moved that tryconsume up into the makeWeaponskill_MCH onConfirm combination so they'll all consume a stack now.

src/Game/Jobs/MCH.ts
AkairyuGestalter: Yeah, I missed it when I was checking things over. Changed both sets of charged actions to default to 3 charges and reduce to 2 when missing the trait

src/Game/Jobs/MCH.ts
AkairyuGestalter: Given how little real-world impact it'll even have on fight timeline planning, I'm going to leave this until and unless the MCH theorycraft people actually ask for it

src/Game/XIVMath.ts
AkairyuGestalter: Sure, hopefully this change makes it a little easier to read! One question I did have while we're talking about this, is the dhRate one checks that the *critBonus* is above 1 in that first ternary check... should it be checking *dhBonus* instead?

src/Components/Jobs/MCH.tsx
AkairyuGestalter: Potentially, but calculating the amount of time remaining would be non-trivial, so I want to wait for user feedback before trying to do so.

src/Game/Constants/MCH.ts
AkairyuGestalter: Not anymore after some back and forth on how to deal with the Queen. Removed.

AkairyuGestalter:
I did double-check snapshotting and monkeying with untargetable markers and everything seems to be behaving the way I'd expect. I may have overlooked some nuance somewhere but it should be good for now.

src/Game/XIVMath.ts
TMTurtle: Good catch, yes that should be dhBonus. It would've never been caught on PCT because hammers auto CDH, there's no button in which the distinction mattered.

PR #71, opened 2024-11-24
additional localization

zqsz-xiv:
Copied suggestions from @UkitaEshiya in this doc: https://docs.google.com/document/d/1cDiR2vR87Mu9Jcwli1rqg46nWE8Yfv7zgV3FRbEJDhs/edit?tab=t.0

Also hooks up partial localization for the "resource overrides" module. Turns out there was more un-localized text than I realized, not sure whether to leave this partially localized or wait until we have translations for all of these buffs.

image


src/Components/PlaybackControl.tsx
```diff
@@ -60,13 +60,21 @@ export function ResourceOverrideDisplay(props: {
 	if (rscInfo.isCoolDown) {
 		str = props.override.type + " full in " + props.override.timeTillFullOrDrop + "s";
 	} else {
-		str = props.override.type;
-		if (props.override.type === ResourceType.LeyLines) str += " (" + (props.override.effectOrTimerEnabled ? "enabled" : "disabled") + ")";
-		if (props.override.type === ResourceType.Enochian) str += " (" + (props.override.effectOrTimerEnabled ? "timer enabled" : "timer disabled") + ")";
-		if (rscInfo.maxValue > 1) str += " (amount: " + props.override.stacks + ")";
+		str = localizeResourceType(props.override.type);
```
zqsz-xiv, line 63: this section needs localization (controls the purple text for set overrides)
image

src/Components/PlaybackControl.tsx
```diff
@@ -973,9 +981,13 @@ export class Config extends React.Component {
 			}
 
 			let timerDesc;
-			if (info.isCoolDown) timerDesc = "Time till full: ";
-			else if (rscType === ResourceType.Polyglot) timerDesc = "Time till next stack: ";
-			else timerDesc = "Time till drop: ";
+			if (info.isCoolDown) {
+				timerDesc = localize({en: "Time till full: "}) as string;
```
zqsz-xiv, line 985: needs localization

src/Components/PlaybackControl.tsx
```diff
@@ -973,9 +981,13 @@ export class Config extends React.Component {
...
+			} else if (rscType === ResourceType.Polyglot) {
+				timerDesc = localize({en: "Time till next stack: "}) as string;
```
zqsz-xiv, line 987: needs localization

src/Components/PlaybackControl.tsx
```diff
@@ -991,7 +1003,7 @@ export class Config extends React.Component {
 
 				{/*stacks*/}
 				<div hidden={!showAmount}>
-					<Input description="Amount: "
+					<Input description={localize({en: "Amount: ", zh: "层数："})}
```
zqsz-xiv, line 1006: this is also used for non-"stack" resources like MP and job gauges, would 数量 be better here?

PR #73, opened 2024-11-24
Predefine remainder of jobs, and default role action skills/resources

AkairyuGestalter:
A smattering of changes I was tinkering with on the side of implementing MCH, but didn't want to include as part of that PR, to make the review passes cleaner:
- Updated the `Common.ts` arrays that define which jobs we support to make it easier for developers to get started on a new job. They now just need to uncomment the job they want to implement from the appropriate role array
- The role arrays also now compose into the all jobs array in the order they appear in the job guide, saving anyone else from needing to sort that out
- favicon switching is now preconfigured for anyone that wants to implement a new job, one less thing for them to think about.
  - We will need to keep the wiki doc for how to get these around somewhere when the next expansion comes along and we need favicons for the new jobs though...
- All role actions, with the exception of the Healer/Caster sleeps and the phys ranged heavy/binds are now defined, and included by default
  - Edit: Now with Limit Breaks too!
- Jobs that aren't fully-implemented yet can still at least display without crashing, and will properly include their respective role actions


AkairyuGestalter:
Rebased to include now-implemented RPR support!

AkairyuGestalter:
Pulled the DoT handling stuff out of this PR since it's getting a much heavier rework with the upcoming BRD PR. This one now focuses on the changes I was making to make it easier for new jobs to get started by including role action and resource definitions by default, as well as making common display logic for those.

src/Components/Jobs/SAM.tsx
```diff
@@ -82,26 +75,10 @@ export class SAMStatusPropsGenerator extends StatusPropsGenerator<SAMState> {
 			makeSamuraiTimer(ResourceType.ThirdEye),
 			makeSamuraiTimer(ResourceType.Tengentsu),
 			makeSamuraiTimer(ResourceType.TengentsusForesight),
-			makeSamuraiTimer(ResourceType.Meditate),
```
zqsz-xiv, line 85: accidental deletion?

src/Controller/Common.ts
```diff
@@ -60,12 +98,30 @@ export const SPS_JOBS = [
 	...HEALER_JOBS, ...CASTER_JOBS
 ]
 
+// Used to default the display of MP and the MP tic timer in the status display
```
zqsz-xiv, line 101:
```suggestion
// Used to default the display of MP and the MP tick timer in the status display
```

src/Controller/Controller.ts
```diff
@@ -736,8 +736,11 @@ class Controller {
 			case ShellJob.WAR:
 				propsGenerator = new WARStatusPropsGenerator(game as WARState);
 				break;
-			default:
+			case ShellJob.BLM:
 				propsGenerator = new BLMStatusPropsGenerator(game as BLMState);
+				break;
+			default:
+				propsGenerator = new StatusPropsGenerator(game);
```
zqsz-xiv, line 743: should we raise an error here instead?

src/Game/GameState.ts
```diff
@@ -765,7 +777,9 @@ export abstract class GameState {
 		let timeTillDamageApplication = 0;
 		if (status.ready()) {
 			if (skill.kind === "spell") {
-				let timeTillCapture = instantCastAvailable ? 0 : (capturedCastTime - GameConfig.getSlidecastWindow(capturedCastTime));
+				let timeTillCapture = instantCastAvailable ? 0 : (capturedCastTime - 
+					(LIMIT_BREAKS.includes(skill.name) ? 0 : GameConfig.getSlidecastWindow(capturedCastTime)) // Can't slidecast an LB
```
zqsz-xiv, line 781: > Can't slidecast an LB

actually, you can

src/Game/Skills.ts
```diff
@@ -303,12 +300,16 @@ export function makeSpell<T extends PlayerState>(jobs: ShellJob | ShellJob[], na
 		potencyFn: (state) => getBasePotency(state, params.potency),
 		jobPotencyModifiers: params.jobPotencyModifiers ?? ((state) => []),
 		validateAttempt: params.validateAttempt ?? ((state) => true),
-		isInstantFn: params.isInstantFn ?? ((state) => true),
+		isInstantFn: params.isInstantFn ?? ((state) => false), // Spells should be assumed to have a cast time unless otherwise specified
```
zqsz-xiv, line 303: I did a quick check and I don't think this affects anything (RDM finishers and PCT holy/comet are still fine). You should also update the comment at the top of `makeSpell` to mention this.

AkairyuGestalter: It was either put it here, or put it in the makeLimitBreak calls into it, and it felt like it technically belonged here more than it did there, since generally speaking a spell should be assumed to have a cast time.

src/Game/GameState.ts
AkairyuGestalter: I guess I should have been more specific, you can't slidecast LB's funky animation lock. This (and a similar change elsewhere), was needed to make sure the full duration of the animation lock didn't allow oGCDs to be used.

src/Game/Jobs/RoleActions.ts
```diff
@@ -203,3 +243,121 @@ makeResourceAbility(ALL_JOBS, SkillName.Sprint, 1, ResourceType.cd_Sprint, {
...
+	// Make the "GCD" appear to roll for the full time of the LB cast and its animation lock
+	// Otherwise the animation lock handling produces a bunch of unexplained empty space in the timeline
+	const recastTime: ResourceCalculationFn<GameState> = (state) => {return params.castTime + state.config.getSkillAnimationLock(name)}
+	const assetName = "Limit Break " + name.charAt(name.length - 1)
+	return makeSpell(jobs, name, unlockLevel, {
+		...params,
+		assetPath: `General/${assetName}.png`,
+		recastTime: recastTime,
+	})
+}
```
misterperson, line 268: Limit Breaks can be woven. E.g. Tank LB1 has low enough animation lock to use like Reprisal without clipping GCD.
Also consider DPS jobs with longer GCDs e.g. Monk Six-Sided Star > LB, Picto swift painting / star prism > LB, etc.

misterperson: I do recall that Blue Mage has a few oGCD spells "Off-guard" and "Peculiar Light" 
image

misterperson: https://www.fflogs.com/reports/rDV3Ytcd9jmgCzRX?fight=1&view=events
image

src/Components/StatusDisplay.tsx
```diff
@@ -557,16 +563,159 @@ export class StatusDisplay extends React.Component {
...
+			timeRemaining: rscCountdown.toFixed(3),
+			className: this.state.hasResourceAvailable(rscType) ? "" : "hidden"
+		}
+	}
+
+	// Jobs should override this to display their enemy-targeted buffs
+	public getJobEnemyBuffViewProps(): BuffProps[] { return [] }
+	
+	// Composes the job-specific buffs with the applicable role buffs
+	public getEnemyBuffViewProps(): BuffProps[] {
```
miyehn, line 589: I've been thinking about renaming this because buffs here aren't necessarily on enemies; just not on self (dance partner for example). But I couldn't come up with a good name ;_;
To make it distinct from `getJobEnemyBuffViewProps`, I guess we can at least rename it to `getAllEnemyBuffViewProps`? (same goes for `getSelfBuffViewProps` and `getResourceViewProps`)

src/Controller/Controller.ts
miyehn: My undersanding is this (as well as a few other changes) makes sure the site doesn't crash when a job is only partially implemented

miyehn (commented):
Regarding changes unrelated to LB:
* I didn't fact check every detail but overall looks good to me. I would appreciate if the nitpicks me and shanzhe commented can go in as well but also not the end of the world otherwise
* Once this goes in can you also update documentation?

LB: I'm not at all familiar with the game mechanics around LB so will leave it to more knowledgeable people,,,
* I'll check back later and see if we should update how LB is visualized though (can be after this PR is in)
  * we do want to figure out if/when movement is allowed around LB (slidecast), so there's the potential to visualize movement on the canvas
* I think it's fine to keep LB in this PR because people already looked at the changes anyway, but it's up to you if you want to split it out xD

src/Controller/Controller.ts
zqsz-xiv: makes sense, a `console.warning` might be helpful though

src/Game/Common.ts
```diff
@@ -69,6 +72,29 @@ enum GeneralSkillName {
 	Never = "Never",
 }
 
+export enum LimitBreakSkillName {
```
zqsz-xiv, line 75: minor detail: do we want to use their proper in-game names ("Shield Wall", "Braver", etc.)? LB3s would take a little bit of extra work, but if we want to match names we should change them now because these actions will get serialized into timelines and be harder to change later.

src/Game/Common.ts
```diff
@@ -94,6 +122,28 @@ export type SkillName = GeneralSkillName
 	| WARSkillName
 ;
 
+export const LIMIT_BREAKS: SkillName[] = [
```
zqsz-xiv, line 125: you should be able to use `Object.values(LimitBreakSkillName) as SkillName[]` (this doesn't work for `const enum` declarations, but should work for ordinary enums)

zqsz-xiv (approved):

miyehn (approved):
^all are minor xD I might fiddle with LB visualization on the timeline after this pr (because its cast bar being on a different horizontal looks a little strange) but the current one works too.

PR #77, opened 2024-11-29
[MCH] Overheated stacks should not be consumed by (most) AoE skills

FloatingGhost:
Per the job guide:

```
Grants 5 stacks of Overheated, each stack allowing the execution of Blazing Shot or Auto Crossbow.
Duration: 10s
Overheated Effect: Increases the potency of single-target weaponskills by 20
Heat Gauge Cost: 50
Overheated effect only applicable to machinist job actions.
```

The only AoE weaponskill to consume stacks of overheated is autocrossbow

this PR adds an `isCleave?` parameter to the `Weaponskill` type, which is used to determine whether we should
eat a stack or not. There was already special-case handling on autocrossbow to eat a stack, which was leading
to it eating 2 stacks per usage. This fixes that too.

### DIFF
--- a/src/Game/Jobs/MCH.ts
+++ b/src/Game/Jobs/MCH.ts
@@ -57,6 +57,14 @@ makeMCHResource(ResourceType.BatteryBonus, 50)
 const COMBO_GCDS: SkillName[] = [SkillName.HeatedCleanShot, SkillName.HeatedSlugShot, SkillName.HeatedSplitShot,
     SkillName.SpreadShot, SkillName.Scattergun // Including AoE GCDs that break the combo, even though they don't combo themselves
 ]
+
+// Skills that don't consume overheat - these are all AoE skills
+// The only AoE skill that consumes overheat is Auto Crossbow
+const WEAPONSKILLS_THAT_DONT_CONSUME_OVERHEAT: SkillName[] = [
+    SkillName.Chainsaw, SkillName.Excavator, SkillName.FullMetalField,
+    SkillName.Scattergun, SkillName.Bioblaster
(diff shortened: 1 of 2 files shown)


src/Game/Jobs/MCH.ts
```diff
@@ -273,7 +274,14 @@ const makeWeaponskill_MCH = (name: SkillName, unlockLevel: number, params: {
                 state.resources.get(ResourceType.WildfireHits).gain(1)
             }
         },
-        (state) => state.tryConsumeResource(ResourceType.Overheated) // All weaponskills executed during overheat will consume a stack
+        // All single-target weaponskills executed during overheat will consume a stack
+        // AoE weaponskills will NOT consume a stack (with the exception of auto crossbow)
+        (state) => {
+            if (!(params.isCleave ?? false)) {
```
zqsz-xiv, line 280: I think it would be better to explicitly list the affected skills (`![SkillName.Chainsaw, SkillName.Excavator, SkillName.FullMetalField, SkillName.SpreadShot, SkillName.Scattergun].includes(name)`) since there's not that many of them, rather than splitting it into a parameter for each skill. This makes it easier to tell if we forgot to list any other skills with this behavior under Overheated.

zqsz-xiv (changes requested):
Thanks for the PR! Left one change request.

FloatingGhost:
updated PR - dropped the `onConfirm` hook from Autocrossbow, hardcoded an array of skills that won't eat a stack

zqsz-xiv (approved):
LGTM!

PR #79, opened 2024-12-02
change the visual of skills on cooldown

Junie-774:
Makes skills on CD resemble how they look in game: with the greyed-out image shrinking as the cooldown progresses. I'd love feedback on both visuals and implementation!

### DIFF
--- a/src/Components/Skills.tsx
+++ b/src/Components/Skills.tsx
@@ -77,6 +77,34 @@ function ProgressCircleDark(props={
 	</svg>
 }
 
+function SecondaryProgressCircle(props={
+	className: "",
+	diameter: 50,
+	progress: 0.7,
+	color: "rgba(0, 0, 0, 0.5)",
+}) {
+	const elemRadius = props.diameter / 2.0;
+	const circleRadius = elemRadius * 0.8;
+	const circumfrence = 2 * Math.PI * circleRadius;
+	const circleFillLength = circumfrence * props.progress;
+	const circleGapLength = circumfrence - circleFillLength;
+	const dashArray = circleFillLength + "," + circleGapLength;
+	const circle = <circle
+		r={circleRadius}
+		cx={elemRadius}
+		cy={elemRadius}
+		fill="none"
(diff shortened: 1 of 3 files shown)


miyehn:
image

Would it be possible to make the progress more obvious? Currently it miiight be hard to tell at first glance, depending on the skill icon :0
But also I don't think we should make the overlay darker... The game solves this by highlighting the edge of the progress. We don't need to replicate exactly but we can probably try adding 2 svg line segments with white drop shadow? Or even also clip it so it's only visible on the brighter side:
image

Would it be worth trying? What do we all think?

--
Also here's something tiny, hopefully it's quick to fix (progress is clipping away part of the proc asset):
image

zqsz-xiv:
+1 to miyehn's comments, in the sample screenshtos it also seems like there's a bit of an optical illusion where the start of the High Thunder + Triplecast lines look a little skewed because the border is near a diagonal line in the base image.

Also, if the ability has available charges, the background shouldn't darken, right?

Junie-774:
clarity: Maybe a small white circle that slightly brightens the image...? kinda like FF does, it might be possible to make it transparent with a white drop shadow to perfectly mimic??? Would that work? Regardless i'll play around with it.

Honestly i'm not sure what the clipping is about, it looks like the circle is in isn't quite centered??? I'm not sure what's going on there but i'll look at it

miyehn:
Clipping: I was referring to the orange dotted lines being thinner at the bottom, where the skill icon is darkened. It happens on Paradox too

Junie-774:
Ok, it should look a decent bit better now. Changed look + fixed clipping + added a secondary CD circle

miyehn:
image
I made some style changes mostly to make things align better. If you don't mind I can push to this branch directly, or if you're still changing things I can hold them until your changes are in first :0

Junie-774:
ooo yeah i like that change. It looks good to me, I'm satisfied with it if you are ^^

miyehn (approved):
🚀

PR #81, opened 2024-12-06
Add prettier formatter

zqsz-xiv:
Adds the [prettier](https://prettier.io/docs/en/) formatter to our codebase. As we get more contributors and more supported jobs, a consistent formatter will help keep code readable. Some files have large diffs because my editor defaults to spaces over tabs, and I wasn't diligent about correcting this when doing a lot of merge work.

I set minimal configuration options in `.prettierrc.toml`. I'll add a CI check for this later, and I'll look to get this merged after the latest batch of jobs (WAR and maybe BRD?) get added to avoid creating merge conflicts for others.


zqsz-xiv:
Added https://www.npmjs.com/package/@yikes2000/prettier-plugin-merge-extras#user-content-return-parentheses to prevent unnecessary wrapping of JSX expressions in parentheses (like `(<div> ... </div>)`).

miyehn:
Can we add `endOfLine = "auto"` to `.prettierrc.toml`? [This option defaults to `lf`](https://prettier.io/docs/en/options#end-of-line) but for people working in Windows this is almost always going to create unnecessarily large diffs (like when I run prettier after pulling this branch). End of line should've be handled by git (or github?) anyway

Also I can live with how prettier handles the `.tsx` files but the parentheses around JSX elements and the extra indentation that comes with them are kinda annoying IMO... I found [this alternative](https://github.com/qpwo/prettier-no-jsx-parens) that looks reasonably up to date but what do you think?

zqsz-xiv:
Changed as requested.

miyehn (approved):
🚀

PR #82, opened 2024-12-08
Make stats tables only include selected nodes when there is selection

miyehn:
* Make stats table only include selected nodes when there is selection
* Make selection summary purple
* When there is only 1 selected node, can now click on it again to unselect it, but still keep the orange cursor at its beginning

image
image

Some entries in the DoT Table (gap/override, total ticks) may not always be meaningful when there is selection. I'm not sure what to do with them and just kept them for now.
image

### DIFF
--- a/src/BlmRotations.test.ts
+++ b/src/BlmRotations.test.ts
@@ -16,12 +16,12 @@
 
 import fs from "node:fs";
 import {controller} from "./Controller/Controller";
-import {TickMode, ShellJob} from "./Controller/Common";
+import {TickMode} from "./Controller/Common";
 import {DEFAULT_BLM_CONFIG, GameConfig} from "./Game/GameConfig";
 import {PotencyModifierType} from "./Game/Potency";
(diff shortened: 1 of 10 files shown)


zqsz-xiv (approved):

zqsz-xiv:
Attaching a before (left)/after (right) comparison for others to see:

image

While testing how this interacts with unapplied skills, I noticed a behavior change in "manual" mode that I think wasn't present before: if at the start of a fight you enter an F3 and then attempt to hit Thunder/F4, the ability will not attempt to be cast. This also exists on live, and I suspect it might have been introduced in #79 with the changes to how cooldown/availability info is passed to the frontend. I'll investigate further once out of FRU prog jail.

zqsz-xiv:
Also, as you can see from the screenshot, dot ticks that occur during the selected duration but were applied by an ability prior to the selection aren't counted. Not really sure what the proper way to handle this is (since we count damage from abilities that haven't yet applied at the end of the selection), so I'm fine with leaving this as-is for now.

PR #83, opened 2024-12-09
Make SkillReadyStatus able to include multiple unavailable reasons

miyehn:
This fixes [this bug](https://github.com/xivintheshell/xivintheshell/pull/82#issuecomment-2526434551), and combines `status` and `nonCdStatus` in `GameState.getSkillAvailabilityStatus`.

While doing the refactor I also removed `validateAttempt` for SAM's meditate because seems like special case handlings elsewhere already achieved the same effect. Lmk if something broke though 🙏

### DIFF
--- a/src/Game/Jobs/SAM.ts
+++ b/src/Game/Jobs/SAM.ts
@@ -1065,7 +1065,7 @@ makeResourceAbility(ShellJob.SAM, SkillName.Meditate, 60, ResourceType.cd_Medita
 	cooldown: 60,
 	applicationDelay: 0.62,
 	// Meditate cannot be used during a GCD roll
-	validateAttempt: (state) => true,
+	validateAttempt: (state) => state.cooldowns.get(ResourceType.cd_GCD).stacksAvailable() > 0,
 	// roll the GCD
 	onConfirm: (state) => {
 		const recastTime = state.config.adjustedSksGCD(
--- a/src/Game/Jobs/SAM.ts
+++ b/src/Game/Jobs/SAM.ts
@@ -1065,7 +1065,7 @@ makeResourceAbility(ShellJob.SAM, SkillName.Meditate, 60, ResourceType.cd_Medita
 	cooldown: 60,
 	applicationDelay: 0.62,
 	// Meditate cannot be used during a GCD roll
-	validateAttempt: (state) => state.cooldowns.get(ResourceType.cd_GCD).stacksAvailable() > 0,
+	validateAttempt: (state) => true,
 	// roll the GCD
 	onConfirm: (state) => {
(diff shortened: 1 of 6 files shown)


src/Game/Jobs/SAM.ts
```diff
@@ -1065,7 +1065,7 @@ makeResourceAbility(ShellJob.SAM, SkillName.Meditate, 60, ResourceType.cd_Medita
 	cooldown: 60,
 	applicationDelay: 0.62,
 	// Meditate cannot be used during a GCD roll
-	validateAttempt: (state) => state.cooldowns.get(ResourceType.cd_GCD).stacksAvailable() > 0,
+	validateAttempt: (state) => true,
```
zqsz-xiv, line 1068: I'd prefer to keep the explicit validation here in case we ever do refactoring that breaks this invariant, or if we ever decide to turn the job logic into a library consumable by frontends other than our own site.

zqsz-xiv (changes requested):
The bug I reported is fixed and Meditate works fine, but it looks like secondary CDs aren't being reflected properly for some abilities. Take MCH's Drill (a GCD with a 20s CD scaled to sks at level 90, or 20s unscaled w/ 2 charges at level 100)--on the live site (right) it grays out while on CD, but on left (PR) it's still lit up. I can look into why when I have some free time, or it might be better if @Junie-774 takes a look first since she implemented this.

image

Junie-774:
the logic for the full-ability grey out is a bit flawed now with the new changes...
I think, the condition should be that stacks < 1 for stacking abilities, and just, 'is on CD' for non-stacking abilites with a secondary cooldown

src/Game/Jobs/SAM.ts
miyehn: Would it make sense if I turn meditate into a skill with secondary cd? I have the intuition that it might let us get rid of some hacks specifically for meditate elsewhere 🤔

zqsz-xiv: We can't directly use a secondary/primary cd here because the GCD roll is scaled to sks and SAM's personal haste buff, while the cooldown field for abilities only takes a scalar. We could turn it into a weaponskill with a secondary cooldown instead, though we'd need to double check to make sure that doesn't break anything.

AkairyuGestalter: It should be fine to define it as a "weaponskill" as far as our code is concerned. DNC does that for Standard and Technical Step.

miyehn: I put its `validateAttempt` back for now. We can change it to a weaponskill in another pr if needed

zqsz-xiv (approved):
🚀

PR #86, opened 2024-12-14
Multi-target support

zqsz-xiv:
Adds basic multi-target support for non-DoT effects by adding a flat multiplier to all potencies in `DamageStatistics.tsx`. Abilities that splash/cleave should specify a `falloff` field representing the _reduction_ in damage to secondary targets, e.g. Flare does 30% reduced damage to other enemies, so it has a `falloff` of 0.3. This forces us to do an extra subtraction when computing multipliers, but makes it easier to copy information from the job guide.

image

TODO:
- other jobs
- (longer-term) figure out how to handle DoTs, dokumori, and chain strat
- clean up target count display on timeline canvas, especially for ogcds


miyehn (changes requested):
We might want to fix a few places in the stats tables :0
image
1. the first high thunder 2 hits 1 target, the second hits 16 targets. But the total potency difference is not reflected in the table (both just show the result of 1 target)
2. the interaction logic is a bit strange here: the 3rd here is not labeled and might look confusing. When I uncheck the 3rd row, the 2nd row is automatically unchecked as well, but if I uncheck the 1st row, the 2nd row stays as-is :0
   * I think we should either 1) combine all instances of high thunder 2 into the first row (plus keep the DoT row) and tell the player to check the below dot table for details, or 2) label each usage including number of targets hit & put the DoT row at the bottom & auto-uncheck the DoT row when any of the previous rows are unchecked

I haven't checked the total potencies in the main table and in the summary section on top, but if there's anything I assume they'll become correct once multitarget + dots are sorted out 🤔

miyehn:
[fight (5).txt](https://github.com/user-attachments/files/18297963/fight.5.txt) <-- the file I used in the above screenshot

zqsz-xiv:
Updated:
image

In the main table: I kept the checkbox and DoT checkboxes right next to each other since the "DoT" portion doesn't correspond to any specific row. As with the other skills in the table, the checkboxes affect all skills of the same name.

I fixed the potency displays in the DoT table. I've chosen not to include a "target count" column since it's kind of crowded already, and you should be able to infer from the potency the number of targets that were hit.

Slightly modified version of your file (I forgot to add a `parseInt` call to the targetCount field, so your version might no longer work)
[aoe_dot_test.txt](https://github.com/user-attachments/files/18305643/aoe_dot_test.txt)

Also, the way the "xN" marker is shown on the timeline currently clips into tick marks for oGCDs. Easiest fix is to make the timeline taller, but that feels inelegant; not sure what the best way to handle this is (maybe moving the text on top of the skill icon instead?)
image

miyehn:
Main table: cool cool by getting rid of the checkbox in the 3rd row it's much more clear. I think we need to add `SkillName.HighThunder2` to `DOT_SKILLS` (in DamageStatistics.ts) as well though.

DoT table: hmm numbers still don't check out :0 in your first screenshot the second high thunder 2 usage should have total potency of 660 + 271.66 * 4 which should be way more than 877.32. Would also expect pot bonus to be more than 5 * 23.6 (because it also gets 1 more tick than above), assuming 23.6 from the row above is correct

"xN" markers: I think the current display is okay because it's at least always readable, or we can move it down just a few more pixels to clip less with the damage marks. Or an alternative idea is something like a triangle background on the top right of skill icon and put the number on top, maybe dark red background with white text? Idk I can mock something up later and see what we prefer-

zqsz-xiv:
>  in your first screenshot the second high thunder 2 usage should have total potency of 660 + 271.66 * 4 which should be way more than 877.32

The "initial" and "DoT" numbers both already include the target count multiplier--you can see the first application on a single target has a per-tick value of 54.33, which multiplied by 5 gives the 271.66 value for 5 targets.

miyehn:
^Yeah but with 4 ticks it should get initial potency on 5 targets (660) + 4 ticks on each of the 5 targets (4 * 271.6) right?

zqsz-xiv:
Oh, I didn't realize you were talking about the total potency column. I'll take a look what the issue is.

zqsz-xiv:
Fixed:
image

miyehn:
image
hmmmm...
image
or top left corner so it's always visible even with icons right after it

I might slightly prefer what you already have until readability becomes a problem but what do you think...

miyehn (approved):
Looks about right now

zqsz-xiv:
Yeah, I think the plain text markers are fine for now. It gets crowded for jobs that have busy weave windows with a lot of cleaves, but it's still understandable if a bit ugly.
image

PR #87, opened 2024-12-14
Update acknowledgements section

miyehn:
Not translated yet but here's a draft for how it could look like in English. Hey y'all let me know your thoughts:
* Content suggestions
* Any name(s) I missed? Hey shanzhe will your friend be maintaining SAM? If not, they should probably at least also be included in the list of additional contributors?
* Fix my broken/awkward English
* By what name(s) would you all like to be listed here? Game/github/discord ID? Something else?

shanzhe you can push to this branch directly. Anyone else let us know below 🙏

### DIFF
--- a/src/Components/IntroSection.tsx
+++ b/src/Components/IntroSection.tsx
@@ -40,59 +40,59 @@ function Changelog() {
 // needs to be a function to evaluate localization
 const getAcknowledgements = () => (
 	<>
-	<div className="paragraph">{localize({en: "This tool is made by:", zh: "作者：", ja: "作者："})}</div>
-	{localize({
-		en: <ul>
-				<li><b>Eshiya (Galahad Donnadieu @ Exodus)</b>: the PM and the big brain BLM</li>
-				<li><b>miyehn (Ellyn Waterford @ Sargatanas)</b>: initial software developer and a humble BLM student</li>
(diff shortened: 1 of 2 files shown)


miyehn:
image
^a screenshot

zqsz-xiv:
> Hey shanzhe will your friend be maintaining SAM? If not, they should probably at least also be included in the list of additional contributors?

I'll take over SAM maintenance since his plan is to retire from raiding after we're done with FRU reclears, but I've added him to the contributors list.

Also, IMO you (@miyehn) should be listed first in the "tool design" + BLM sections, like how academic papers list the most important author first, since you built the whole thing initially. I'd feel weird about my name being first for both since I joined the project relatively recently, and most of my work for BLM was refactoring.

miyehn:
> you should be listed first in the "tool design" + BLM sections

I already listed myself at the top for making BLM in the Shell during EW. You were the one pushing the tool forward since DT and at this point I think you are more fit to be considered the primary maintainer. <-- unless you don't want to :0
I'm still around mostly to make sure it doesn't go off the rails - or let that be my goal. ~~I've been trying to shift focus onto my other projects but haven't been as successful😂~~

zqsz-xiv:
Sure, if that's the way you feel about it then we can keep what you have now.

PR #88, opened 2024-12-19
BRD in the shell... and one hell of a DoT overhaul

AkairyuGestalter:
Like it says in the title, this PR _**HEAVILY**_ refactors how DoTs are tracked and processed. Most of the logic has been pushed down into the core game state tracking, with functions for jobs to call to hook into that processing. End user visible stuff that has changed re: DoTs:
- BLM now distinguishes between the four types of Thunder DoT effects, but treats them all as a single unit for the uptime tracking information in the summary
- SAM now shows Higanbana DoT uptime info in the summary
- MCH does *not* show Bioblaster uptime
- Damage markers on the timeline will display all damaging effects that applied at a given timestamp, so multiple DoTs can be displayed (as is the case on BRD)

A couple additional refactoring efforts I made to avoid perpetuating problems as I implented BRD behavior:
- Moved animation delay overrides to the action definitions, instead of a GameConfig method
  - movement skills now select their animation lock by importing the MOVEMENT_SKILL_ANIMATION_LOCK constant from `Skills.ts`
  - Shield pop skills can set their fake animation lock when calling `makeAbility`
- Moved buff cover logic to job override methods instead of requiring core GameState changes
- Deduplicated party buff icon files where jobs have added them now
  - Fixed RDM's Embolden buff icons, there's one that goes on the RDM and one that goes on the Party, and they were unintentionally swapped


src/Components/Localization.tsx
```diff
@@ -354,7 +354,10 @@ const resourcesZh = new Map<ResourceType, string>([
 	[ResourceType.Triplecast, "三重咏唱"],
 	[ResourceType.Firestarter, "火苗"],
 	[ResourceType.Thunderhead, "雷砧"],
-	[ResourceType.ThunderDoT, "高闪雷"],
+	[ResourceType.HighThunder, "高闪雷"], // May need retranslation from ThunderDoT
+	[ResourceType.ThunderIII, "高闪雷"], // May need retranslation from ThunderDoT
+	[ResourceType.HighThunderII, "高闪雷"], // May need retranslation from ThunderDoT
```
zqsz-xiv, line 359:
```suggestion
	[ResourceType.HighThunderII, "高震雷"], // May need retranslation from ThunderDoT
```

src/Components/Localization.tsx
```diff
@@ -354,7 +354,10 @@ const resourcesZh = new Map<ResourceType, string>([
...
+	[ResourceType.ThunderIV, "高闪雷"], // May need retranslation from ThunderDoT
```
zqsz-xiv, line 360:
```suggestion
	[ResourceType.ThunderIV, "霹雷"], // May need retranslation from ThunderDoT
```

src/Components/Localization.tsx
```diff
@@ -354,7 +354,10 @@ const resourcesZh = new Map<ResourceType, string>([
 	[ResourceType.Triplecast, "三重咏唱"],
 	[ResourceType.Firestarter, "火苗"],
 	[ResourceType.Thunderhead, "雷砧"],
-	[ResourceType.ThunderDoT, "高闪雷"],
+	[ResourceType.HighThunder, "高闪雷"], // May need retranslation from ThunderDoT
+	[ResourceType.ThunderIII, "高闪雷"], // May need retranslation from ThunderDoT
```
zqsz-xiv, line 358:
```suggestion
	[ResourceType.ThunderIII, "暴雷"], // May need retranslation from ThunderDoT
```
Grabbed some translations from the Chinese [job guide](https://actff1.web.sdo.com/project/20190917jobguid/index.html#/continfo/blackmage/pve).

src/Components/TimelineCanvas.tsx
```diff
@@ -445,30 +445,38 @@ function drawDamageMarks(
...
-			info = potency.toFixed(2) + " (" + sourceStr + ")";
-			if (pot) buffImages.push(buffIconImages.get(BuffType.Tincture));
-
-			dm.potency.getPartyBuffs().forEach((desc) => {
-				buffImages.push(buffIconImages.get(desc));
-			});
-		}
+		const info: string[] = [];
+		let buffImages: Array<HTMLImageElement | undefined> = [];
+		dm.damageInfos.forEach((damageInfo) => {
```
zqsz-xiv, line 450: how come this has changed into a `for` loop? is this to accomodate multiple damage events on a single marker (like simultaneous dot ticks)?

src/Controller/DamageStatistics.ts
```diff
@@ -91,48 +93,26 @@ function expandDoTNode(node: ActionNode, lastNode?: ActionNode) {
...
-	for (let i = 0; i < node.getPotencies().length; i++) {
-		if (i > 0) {
-			let p = node.getPotencies()[i];
-			if (p.hasResolved()) {
-				entry.totalNumTicks += 1;
-				entry.baseDotPotency = p.base;
-				if (p.hasHitBoss(bossIsUntargetable)) {
-					entry.numHitTicks += 1;
-				}
+	for (let i = 0; i < node.getDotPotencies(dotName).length; i++) {
```
zqsz-xiv, line 109: I think this used to start at i=1 because to skip the potency of application, did this change?

Also, you may as well just iterate over `node.getDotPotencies(dotName)` directly now.

src/Game/Jobs/BLM.ts
```diff
@@ -1103,14 +1049,10 @@ makeSpell_BLM(SkillName.HighThunder, 92, {
 	baseCastTime: 0,
 	baseManaCost: 0,
 	basePotency: 150,
-	applicationDelay: 0.757,
+	applicationDelay: 0.76,
```
zqsz-xiv, line 1052: I think the 0.757 number is more precise and measured from logs, so we should leave this as-is.

src/Game/XIVMath.ts
```diff
@@ -142,6 +143,24 @@ export class XIVMath {
 		if (buff === ResourceType.Fuka) {
 			return Traits.hasUnlocked(TraitName.EnhancedFugetsuAndFuka, level) ? 13 : 10;
 		}
+		if (buff === ResourceType.ArmysPaeon) {
+			const repertoire = controller.game.resources
```
zqsz-xiv, line 147: I'd prefer to minimize accesses to the global `controller` object, would it be possible to set up the different levels of Army's Paeon/Muse as distinct buffs that we pass in?

src/Game/Jobs/RDM.ts
```diff
@@ -123,6 +134,21 @@ export class RDMState extends GameState {
 		this.registerRecurringEvents();
 	}
 
+	override jobSpecificAddDamageBuffCovers(node: ActionNode, skill: Skill<PlayerState>): void {
+		if (this.hasResourceAvailable(ResourceType.Embolden) && skill.aspect !== Aspect.Physical) {
+			node.addBuff(BuffType.Embolden);
+		}
+		if (this.hasResourceAvailable(ResourceType.Manafication)) {
```
zqsz-xiv, line 141:
```suggestion
		if (this.hasResourceAvailable(ResourceType.Manafication) && skill.kind === "spell" || skill.kind === "weaponskill") {
```
manafic doesn't affect ogcds

src/Game/Jobs/PCT.ts
```diff
@@ -1129,6 +1166,7 @@ makeAbility_PCT(SkillName.TemperaCoatPop, 10, ResourceType.cd_TemperaPop, {
 	],
 	startOnHotbar: false,
 	applicationDelay: 0,
+	animationLock: 0.01,
```
zqsz-xiv, line 1169: can you make the 0.01 into something like `FAKE_POP_SKILL_ANIMATION_LOCK`?

zqsz-xiv (changes requested):
Thanks, everything looks a lot cleaner than before.

A couple bugs I noticed:
- This timeline is raising `Controller.ts:745 Assertion failed: Reported dropping High Thunder II when no coverage was detected`
[2dot_assertion_error.txt](https://github.com/user-attachments/files/18207031/2dot_assertion_error.txt)

- I tried cross-checking with one of the TOP timelines you uploaded a while ago, and the site's freaking out when trying to report gaps for dot applications that have clips.
[akairyu_BLM-TOP-2284_Sps-Dawntrail_Launch.txt](https://github.com/user-attachments/files/18207046/akairyu_BLM-TOP-2284_Sps-Dawntrail_Launch.txt)

Also, would you mind adding some automated tests for this? There's some edge cases that are hard to test manually (especially with BLM's mutually exclusive dots), and some sanity checks on the dot table output would be useful.

miyehn:
(haven't looked at that much code yet but)
just noticed there are at least two buff covers with almost the same color and I wonder if we can do something to differentiate them :0
image

Also looks like the BRD standard opener starts like this (see below). But because Wanderer's Minuet can only be used in combat, currently in the tool it is not possible to double weave without clipping. Has this not been a problem for other jobs with sckills that require in combat? :0
image

src/Components/TimelineCanvas.tsx
AkairyuGestalter: That's exactly the reason. BRD especially will have this a lot, since both Caustic Bite and Stormbite will be ticking at the same time, but it also accommodates a DoT ticking and a GCD/oGCD landing at exactly the same time as well.

src/Controller/DamageStatistics.ts
AkairyuGestalter: The initial potency of the node is now fully separate from the node's DoT potencies, in both storage and access, so it didn't need to iterate from 1 anymore, but you're right that we can just foreach now. Done.

src/Game/Jobs/BLM.ts
AkairyuGestalter: I don't remember intentionally changing this, so sure, back to the more precise value it goes!

src/Game/XIVMath.ts
AkairyuGestalter: I'm going to punt on this one for now. I have an alternative that I'm prototyping at the moment to move from using string enums to define so much of our data to using keyed interfaces instead, with the original intent of getting us towards being able to resolve the issue of exporting action IDs to Amarantine instead of the names. Doing so would also give us the opportunity to pass data like this around more easily. I hate the idea of needing to define and gain/force drop something like eight different flavors of the same buff (I already hate needing to have three different Radiant Finales for a similar reason).

src/Game/Jobs/PCT.ts
AkairyuGestalter: Sure thing, and since you pointed out the three decimal precision of the other thing, I'm also going to make this be 0.001s to really drive home the "it ain't a real lock" behavior!

AkairyuGestalter:
Functional issues should be addressed now, I have not yet looked at adding tests for any of this, but will try to give that a spin this weekend.

src/Game/Jobs/PCT.ts
zqsz-xiv: I think that makes sense, but I'm a worried it might break existing timelines that operate with the 0.01s "animation lock" assumption.

src/Components/TimelineCanvas.tsx
zqsz-xiv: Makes sense, I think I'll hold the multi-target PR for now so I can make use of this change.

miyehn:
Suggesting a few colors for BRD resources, so the repertoire stacks aren't black/white:
* `pitchPerfect: "#9ac5f1"`
* `repertoire: "#e8c35a"`

^for both color themes

image
image

src/Game/XIVMath.ts
miyehn: I'm going to second the idea of moving controller out of this file 🤔
XIVMath should really just be a library of math formulas at the very bottom of the include tree and shouldn't depend on any outside state. Even if there's going to be more refactors later, as a quick fix for now can we pass in `repertoire` and `museRepertoire` as arguments? And/or even default them to 0 so it doesn't affect other jobs calling this function

src/Game/Jobs/PCT.ts
AkairyuGestalter: ooh, hadn't thought about that... I think it's unlikely to actually break anything but probably better to play it save. Will return it to 0.01.

src/Game/XIVMath.ts
AkairyuGestalter: Changed the XivMath functions to just receive the integer modifier that we were previously calculating in XivMath itself, and moved the calculation functions into job modules instead. Can polish further later as we go.

zqsz-xiv (commented):
Code changes look good to me, but I'm still seeing issues with reported DoT gaps. Here's what the thunder table for your TOP timeline looks like:

image

You can also see this in this more minimal reproduction. Note also that there are 8 thunder casts, but only 6 rows in the table.
[spam_dot.txt](https://github.com/user-attachments/files/18220193/spam_dot.txt)

src/Game/Skills.ts
```diff
@@ -466,12 +489,22 @@ export function makeAbility<T extends PlayerState>(
 	if (!Array.isArray(jobs)) {
 		jobs = [jobs];
 	}
+	const onApplication: EffectFn<T> = combineEffects(
+		(state, node) => (node.applicationTime = state.time),
+		params.onApplication ?? NO_EFFECT,
+	);
+	// All abilities that require being in combat should check isInCombat
+	const validateAttempt: StatePredicate<T> = combinePredicatesAnd(
+		(state) => (params.requiresCombat ? state.isInCombat() : true),
```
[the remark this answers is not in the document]
AkairyuGestalter: Trying this suggestion broke a bunch of unit tests, so either we leave it for now, or try a different polish method

AkairyuGestalter:
Urgh, I was so focused on fixing the console errors I didn't even notice the stupid in the table itself. I believe I have squashed the remaining bugs, and the revised BLM DoT test should cover the lowest level version of the problems I found.

Note that the dotspam file needs to advance past the last T3's application before it shows in the table :)

src/Game/Skills.ts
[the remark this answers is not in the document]
zqsz-xiv: Oops, I meant `!params.requiresCombat || state.isInCombat()` (missed a !).

zqsz-xiv (approved):
Looks good, it looks like this actually incidentally fixes some bugs in gap + override calculation found on live, where it seems like T3 is still treated as 30s for gap calculation, and the total gap on live also seems to be almost the whole duration of the fight for some reason.

One more edge case with gap/override (I'm ok with merging without fixing this if it's too much effort to fix):
[2dot_needs_bigger_gap.txt](https://github.com/user-attachments/files/18220726/2dot_needs_bigger_gap.txt)
The HT2 overrides the initial HT application, but the dot table is unaware of this when computing gap/override for the next application.

AkairyuGestalter:
That 2dot bigger gap file looks correct to me, it's just a change in expectation since HT and HT2 are being treated as the same effect for the purposes of gap/overwrite calculations:
- The HT2 applies at 1.4 and lasts 24s until 25.4
- The second HT applies at 33.757
- The gap between these two is 8.357, and its correctly attributed to the second HT's application

I'll stick a changelog in after raid :+1: then we'll be good to merge!

zqsz-xiv:
Oh, I didn't realize that the "gap" accounted for the HT2's application as well--I guess this makes sense, given that the override duration of the first HT seems to be attributed to the HT2 table. LGTM in that case.

PR #90, opened 2024-12-20
Localization for the recently added jobs

miyehn:
Crowdsourcing translations here >:D
【腾讯文档】排轴器职业汉化
https://docs.qq.com/sheet/DR1hLdUFaR0xranh1?tab=BB08J2

Will continue working here and rebase & merge after BRD is in


miyehn:
this branch now contains everything from #94 and should be ready to go once that one is in

miyehn:
I still need to test the changes more, probably after #106. But should otherwise be ready

zqsz-xiv:
I saw that you deleted your script since it no longer works with the new refactor, but do you think it would be worth keeping around in case we need to do another large localization update like this? I'm ambivalent about whether to keep the source of truth for translations within our codebase or an external spreadsheet, but in the long term if we add more jobs (or official jp support) it may be better to maintain a qq/google sheet and pull text updates from there.

src/Components/DamageStatistics.tsx
```diff
@@ -952,9 +743,6 @@ export class DamageStatistics extends React.Component {
 			// tags
 			let tags: React.ReactNode[] = [];
 			tags.push(props.row.displayedModifiers.map((tag, i) => <BuffTag key={i} buff={tag} />));
-			if (props.row.baseMainPotency === 400) {
-				tags.push(<BuffTag key={tags.length} tc={true} />);
-			}
```
zqsz-xiv, line 957: rip thundercloud o7

src/Game/Data/Jobs/MCH.ts
```diff
@@ -71,8 +71,8 @@ export const MCH_STATUSES = ensureRecord<ResourceData>()({
 	WILDFIRE: { name: "Wildfire", label: { zh: "野火（敌）" } },
 	WILDFIRE_SELF: { name: "Wildfire Self", label: { zh: "野火（我）" } },
 	FLAMETHROWER: { name: "Flamethrower", label: { zh: "火焰喷射器持续中" } },
-	BIOBLASTER: { name: "Bioblaster", label: { zh: "毒菌冲击" } },
-	TACTICIAN: { name: "Tactician", label: { zh: "策动" } },
+	BIOBLASTER: { name: "Bioblaster", label: { zh: "毒菌冲击dot" } },
+	TACTICIAN: { name: "Tactician", label: { zh: "策动buff" } },
```
zqsz-xiv, line 75: Do these need the "buff"/"dot" suffixes?

src/Components/ColorTheme.tsx
```diff
@@ -420,6 +421,58 @@ export let getCurrentThemeColors: () => ThemeColors = () => {
...
+		[PotencyModifierType.SURGING_TEMPEST, MarkerColor.Purple],
+		[PotencyModifierType.BARRAGE, colors.brd.barrage],
+		[PotencyModifierType.RAGING_STRIKES, colors.brd.ragingStrikes],
+		[PotencyModifierType.BATTLE_VOICE, colors.brd.battleVoice],
+		[PotencyModifierType.RADIANT_FINALE_THREE_CODA, colors.brd.radiantFinale],
+		[PotencyModifierType.RADIANT_FINALE_TWO_CODA, colors.brd.radiantFinale],
+		[PotencyModifierType.RADIANT_FINALE_ONE_CODA, colors.brd.radiantFinale],
+		[PotencyModifierType.WANDERERS_MINUET, colors.brd.wanderersCoda],
+		[PotencyModifierType.MAGES_BALLAD, colors.brd.magesCoda],
+		[PotencyModifierType.ARMYS_PAEON, colors.brd.armysCoda],
```
zqsz-xiv, line 467: needs NO_MERCY from GNB

src/Game/Data/Jobs/MCH.ts
miyehn: Not mandatory but I think it is preferred here to differentiate between these resources and the skills with the same name. 
These two suffixes were there in the player-filled-out sheet before and I removed them at some point, before I realized why they were there (when I saw the sheet for WAR) :0
It's not super consistent between jobs but at least they are from people who actually play the jobs xD

zqsz-xiv: I think the text here would only show up when hovering a buff icon (in which case it's self-evident that it's a buff/dot), or in the resource override dropdown (where it might be less obvious). My preference is to avoid the suffix for consistency with the en version of the site since we don't make the distinction there, and for consistency with other jobs.

miyehn:
> do you think it would be worth keeping around in case we need to do another large localization update like this?

Nah I can just make another script.

> whether to keep the source of truth for translations within our codebase or an external spreadsheet

Currently, code should be the source of truth, everything in the sheet is just for communication convenience and I'll ask people to doublecheck the site once changes here go live anyway.

Unless we want to separate localization data from code at some point- I personally don't think it's necessary for Chinese and might just slow down development, but I'm not thinking super long term and/or for other languages.. probably your opinion matters more here now that you define our scope xD

src/Game/Data/Jobs/MCH.ts
miyehn: Hmm I agree with the consistency issue but I'm still going to keep them for now, but I'll bring it up to those who filled out the sheet once it's live and see what they think.
The translation is mostly for the resource overide dropdown. It also shows up when hovering a buff icon (a bit redundant) but I don't think people read that hover text very often anyway.
Also in the Chinese player community it is much more common to refer to skills/buffs/resources by shorter, more memorable nicknames, and over time people might completely forget about their official names (you'll be amazed how many black mages don't know the official skill names lol). So I think that's where these suffixes can help.

zqsz-xiv:
@miyehn Anything else you wanted to add, or is this ok to merge?

miyehn:
Sorry I was busy irl... I'm half way through checking it with your transcribed lines; I'll probably finish up later tonight and merge

PR #91, opened 2024-12-20
Remove flank/rear positionals for non-melee jobs

miyehn:


src/Components/StatusDisplay.tsx
```diff
@@ -762,26 +760,6 @@ export class StatusPropsGenerator<T extends PlayerState> {
 			this.makeCommonTimer(ResourceType.Sprint),
 		);
 
-		// Melee jobs should end with the "I am able to hit this positional" selectors
-		if (MELEE_JOBS.includes(job)) {
```
AkairyuGestalter, line 766: Unless I'm missing something, this change will mean no melee job gets the "status" display of the positionals that shanzhe set up to control whether you could hit that positional or not at that time. I think we still need this

miyehn: Oh rip lmaooo I was dumb (just put them back)

PR #93, opened 2024-12-23
MCH - Add Flamethrower

AkairyuGestalter:
Flamethrower has it's use-case now: FRU phase 4 optimization. It has the following behavior characteristics:
- Channeled skill like Meditate and Improvisation: all other actions cancel it
- Technically a ground-targeted DoT: It gets a special DoT tic on application
- Flamethrower tics once per second, rather than once every three

Core changes to support this:
- DoT registration with `registerRecurringEffects` can now specify that the core DoT tick handling will exclude that effect (Job code will be expected to deal with it)
- DoT registration can now specify that the DoT is a ground-targeted DoT
  - When addDotPotencies is called for a ground-targeted DoT, an extra potency will be added to the node for the special onApplication tick
  - When applyDot is called for a ground-targeted DoT, the first tick will be immediately handled
- DoT potency creation can now optionally specify the frequency at which the DoT ticks. The number of potencies added to the node will respect this frequency if provided.
- DoT tick tracking in the controller can now track ticks for specific DoTs in addition to the standard tick, to deal with the ground target initial ticks and specially-handled DoTs.
  - When retrieving the maximum number of potential ticks for a DoT, it will combine the specific ticks for that DoT with the total number of standard ticks, unless told to include only the DoT-specific ticks.

I also updated the damage statistics table to hide the total ticks denominator for DoTs that are not expected to be kept up full time, since there will always be many more potential ticks than actual ticks.

### DIFF
--- a/src/Components/DamageStatistics.tsx
+++ b/src/Components/DamageStatistics.tsx
@@ -1231,7 +1231,7 @@ export class DamageStatistics extends React.Component {
 						<div style={cell(20)} />
 						<div style={cell(8)}>
 							{
-								/* The total dick denominator isn't terribly useful for DoTs that aren't maintained full-time */
+								/* The total tick denominator isn't terribly useful for DoTs that aren't maintained full-time */
 								controller.game.fullTimeDoTs.includes(dotTable.dotName) ? (
 									<>
 										{dotTableSummary.totalTicks}/{dotTableSummary.maxTicks}
--- a/src/Components/DamageStatistics.tsx
+++ b/src/Components/DamageStatistics.tsx
@@ -1230,7 +1230,16 @@ export class DamageStatistics extends React.Component {
 						<div style={cell(10)}>{dotTableSummary.cumulativeOverride.toFixed(3)}</div>
 						<div style={cell(20)} />
 						<div style={cell(8)}>
-							{dotTableSummary.totalTicks}/{dotTableSummary.maxTicks}
+							{
+								/* The total dick denominator isn't terribly useful for DoTs that aren't maintained full-time */
+								controller.game.fullTimeDoTs.includes(dotTable.dotName) ? (
+									<>
+										{dotTableSummary.totalTicks}/{dotTableSummary.maxTicks}
+									</>
+								) : (
+									<>{dotTableSummary.totalTicks}</>
+								)
+							}
 						</div>
 						<div style={cell(24)}>
 							{dotTableSummary.totalPotencyWithoutPot.toFixed(2)}
--- a/src/Game/Jobs/RoleActions.ts
+++ b/src/Game/Jobs/RoleActions.ts
@@ -8,19 +8,10 @@ import {
 	TANK_JOBS,
 } from "../../Controller/Common";
 import { SkillName, ResourceType, TraitName, WarningType, TankLBResourceType } from "../Common";
-import {
-	combineEffects,
-	makeAbility,
-	makeLimitBreak,
-	makeResourceAbility,
-	makeSpell,
-} from "../Skills";
+import { makeAbility, makeLimitBreak, makeResourceAbility, makeSpell } from "../Skills";
 import { DoTBuff, EventTag, makeResource } from "../Resources";
 import type { GameState } from "../GameState";
 import { controller } from "../../Controller/Controller";
-import { SAMState } from "./SAM";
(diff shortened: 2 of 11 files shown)


src/Components/DamageStatistics.tsx
```diff
@@ -1230,7 +1230,16 @@ export class DamageStatistics extends React.Component {
 						<div style={cell(10)}>{dotTableSummary.cumulativeOverride.toFixed(3)}</div>
 						<div style={cell(20)} />
 						<div style={cell(8)}>
-							{dotTableSummary.totalTicks}/{dotTableSummary.maxTicks}
+							{
+								/* The total dick denominator isn't terribly useful for DoTs that aren't maintained full-time */
```
zqsz-xiv, line 1234: :huh:
```suggestion
								/* The total tick denominator isn't terribly useful for DoTs that aren't maintained full-time */
```

src/Game/Jobs/RoleActions.ts
```diff
@@ -46,6 +47,12 @@ const cancelImprovisation = (state: GameState) => {
 	}
 };
 
+const cancelFlamethrower = (state: GameState) => {
+	if (state.job === ShellJob.MCH) {
```
zqsz-xiv, line 51: Since these are guarded by `state.job`, maybe we should combine meditate + improv + flamethrower into a single `cancelChannel` function so we don't have to repeat `combineEffects` everywhere.

zqsz-xiv (changes requested):
Getting a crash when trying to load your TOP timeline (in fact, it caused a crash on page load since I already had it saved):
```
Cannot read properties of undefined (reading 'removeUnresolvedDoTPotencies')
TypeError: Cannot read properties of undefined (reading 'removeUnresolvedDoTPotencies')
    at BLMState.applyDoT (http://localhost:3000/static/js/bundle.js:24077:20)
    at onApplication (http://localhost:3000/static/js/bundle.js:24968:41)
    at Object.onApplication (http://localhost:3000/static/js/bundle.js:33791:7)
    at Event.effectFn (http://localhost:3000/static/js/bundle.js:23811:15)
    at http://localhost:3000/static/js/bundle.js:23652:15
    at Array.forEach (<anonymous>)
    at BLMState.tick (http://localhost:3000/static/js/bundle.js:23647:26)
    at Controller._requestTick2 (http://localhost:3000/static/js/bundle.js:18938:32)
    at Controller._replay2 (http://localhost:3000/static/js/bundle.js:19126:197)
    at Controller.loadBattleRecordFromFile (http://localhost:3000/static/js/bundle.js:18236:200)
```

I'll cross-check the damage tables with some more test records once this is fixed.

zqsz-xiv:
Also, should we hide the `gap` + `override` columns as well for dots like Flamethrower that aren't meant to be full uptime?

src/Game/Jobs/RoleActions.ts
AkairyuGestalter: I did briefly think about that, and I think this is a good idea. I'll get that change made 👍

AkairyuGestalter:
Re: gap/override. I didn't want to mess with the overall formatting of the tables, and the gap output might still be useful for DoTs that you're expected to keep on cooldown, but not full uptime (thinking Salted on DRK, GNB's two burst DoTs) to see if you're drifting uses later in the fight.

And ofc I didn't think to try loading what should have been a completely unrelated fight timeline in with these changes 😑 will see what stupid I have done to myself _this time_.

src/Game/Jobs/RoleActions.ts
AkairyuGestalter: Channeled skill cancelling is now pushed down into core GameState and Skills handling that jobs can override opt into. I did take the fake pop skill stuff into account when doing this, so those should not result in a channel cancel. No more need for special case handling all over the place!

AkairyuGestalter:
The TOP timeline crash turned out to be due to a half-baked change I'd made and subsequently forgotten about when I was working on handling the "ground target initial tick" behavior. Reverting that restored that to functionality.

While testing the various channel cancelling cases, I also found a couple more actions in MCH and DNC (Barrel Stabilizer and Flourish) that didn't have the newer `requiresCombat` configuration for auto advancing to the first damage application, so I went ahead and fixed those too.

zqsz-xiv (approved):
🚀

Issue #78, opened 2024-12-02
Export skill ID instead of skill name for CSV for Tischel's plugin


miyehn:
So Chinese players can also use that plugin for practice. 

Blizzard in Cyan is 34653: https://ff14.huijiwiki.com/wiki/Action:34653

The question being what's the fastest way to associate skills with their IDs- probably shouldn't find out by searching every skill in huijiwiki.

zqsz-xiv:
It looks like the numbers are in the URLs of the Garland DB links: https://garlandtools.org/db/#action/34653 (blizzard in cyan)

There might be an API to enumerate them.

AkairyuGestalter:
I've not spent much time looking to see if there's a good search method on xivapi, to retrieve a list of all actions and their IDs for a given job, but https://arcanedisgea.github.io/xiv-id is an easy way to look them up one by one (or just crib from xivanalysis' codebase...)

zqsz-xiv:
> or just crib from xivanalysis' codebase...
@AkairyuGestalter are you referring to these files? https://github.com/xivanalysis/xivanalysis/blob/def477cc6fce5571604cb0f6968d313dcedf6a3e/src/data/ACTIONS/root/AST.ts

Easiest thing for us to do might be just writing a script to extract the `id` from all of these files, and then manually copy them into the declarations for our already-implemented jobs.

zqsz-xiv:
Done as of a long time ago (#94, #339, etc.).

PR #94, opened 2024-12-24
Convert most data from enums to keyed interfaces

AkairyuGestalter:
This PR proposes moving most of our string enum data definitions to a set of keyed interfaces, which can allow us to expand on the data offered beyond what enums allow. The sets of data I'm moving as of this PR:
- Job and Role definitions
- Action (ie SkillNames)
- Resources and Cooldowns
- Traits

A few things this allowed me to do was to remove some use of arbitrary arrays controlling things like:
- Does this job use MP?
- Is this resource toggle-able in the status display?
- Can this resource be clicked off in the status display?
- How many stacks does this status effect resource have?
  - This also now drives the stacked icon registration behavior

In addition, for these sets of Data, I moved the source of the translated strings out of maps in `Localization.tsx` to the Data definitions themselves.

I also provided a few examples of unused properties that we might eventually move to using for future updates such as:
- Action ID: For exporting to Tischel's plugin and/or other systems, related to Issue #78 
- Cooldown maximum, and Cooldown charges
  - We could move to getting more data from the interface type when building out job behavior code

I considered but did not include moving the BuffType enum used for the skill timeline buff covers and party buff markers, and the WarningType enum to this format. Doing so will be left for future work.

In addition to the above, as part of the ongoing effort to consolidate where developers need to look to add their new jobs to the site, I moved the function that returns the job-specific GameState class out of `Controller.ts` and into the `Game/Jobs/index.ts` file, and pushed the retrieval of the job-specific StatusPropsGenerator UI class into a getter on the `GameState` class. This will mean in most cases, job developers don't need to jump over to the `Controller` class to hook up their job, it'll be done in or near files they'd be editing anyways.


AkairyuGestalter:
Added traits port for a further example of usage

AkairyuGestalter:
- The type checker blocks on strings not in the key set (as long as it knows it's expecting a ShellJob key, ActionKey, etc). That was the main reason I pulled over the `ensureRecord` utility from xivanalysis instead of just exporting a bunch of `Record<string, T>` objects, was to get that type checking on the keys.
- The plethora of blank job files under the `Data/<etc>/Jobs` directories are indeed things I just haven't ported, or stub files for jobs nobody has implemented yet to make it clearer where their data should go. It's an extension of the "predefine as much as possible" PR I had a whole back, I want to make it easier for a new job contributor to find where things need to go, without needing to do the work of stitching it all together.

miyehn:
I'll probably also take a look at it starting 1/2/25, after I get back to GMT-8

zqsz-xiv (commented):
I'll take a closer look and do some testing over the weekend, but wow: things definitely look cleaner after this PR.

One question: would it be possible to combine actions/resources/cooldowns for each class into a single file? It would give a little less overhead to job implementors.

AkairyuGestalter:
> I'll take a closer look and do some testing over the weekend, but wow: things definitely look cleaner after this PR.
> 
> One question: would it be possible to combine actions/resources/cooldowns for each class into a single file? It would give a little less overhead to job implementors.

Yeah, I can crush the different sets of data down into a single file per job to make it even easier to find. Will sort that out tomorrow!

src/Game/Data/Jobs.ts
```diff
@@ -0,0 +1,246 @@
...
+		label: localize({ en: "Caster" }),
+	},
+});
+export type RoleType = typeof ROLES;
+export type RoleKey = keyof RoleType;
+
+export interface Job {
+	role: RoleKey;
+	implementationLevel: ImplementationKey;
+	speedStat: "sks" | "sps";
```
zqsz-xiv, line 57: Just to make sure, is this only used for calculating DoT scaling? Can we just infer this from `role` instead?

src/Game/Data/Jobs/BLM.ts
```diff
@@ -0,0 +1,119 @@
...
+	cd_MANAFONT: { name: "cd_Manafont" }, // [0, 1x]
+	cd_AMPLIFIER: { name: "cd_Amplifier" }, // [0, 1x]
+	cd_RETRACE: { name: "cd_Retrace" }, // [0, 1x]
+});
+
+export const BLM_GAUGES = ensureRecord<Resource>()({
+	POLYGLOT: { name: "Polyglot", label: { zh: "通晓" } }, // [0, 3]
+	ASTRAL_FIRE: { name: "AstralFire", label: { zh: "星极火" } }, // [0, 3]
+	UMBRAL_ICE: { name: "UmbralIce", label: { zh: "灵极冰" } }, // [0, 3]
+	UMBRAL_HEART: { name: "UmbralHeart", label: { zh: "冰针" } }, // [0, 3]
```
zqsz-xiv, line 60:
```suggestion
	ASTRAL_FIRE: { name: "Astral Fire", label: { zh: "星极火" } }, // [0, 3]
	UMBRAL_ICE: { name: "Umbral Ice", label: { zh: "灵极冰" } }, // [0, 3]
	UMBRAL_HEART: { name: "Umbral Heart", label: { zh: "冰针" } }, // [0, 3]
```

zqsz-xiv: are these text declarations currently used anywhere?

src/Game/Data/types.ts
```diff
@@ -0,0 +1,34 @@
+import { LocalizedContent } from "../../Components/Localization";
+
+// Properties beyond name are all optional at this time to allow for gradual migration
+// Mainly provided here as examples of what we could do with the data
+
+export interface Action {
+	name: string;
+	id?: number; // Intended for use with exporting to Amarantine
```
zqsz-xiv, line 8:
```suggestion
	id?: number; // Intended for use with exporting to Tischel
```
this is the in-game ID, correct?

src/Utilities/ensureRecord.ts
zqsz-xiv, line 1: Can we just have a single top-level `src/utilities.ts` file instead of a separate file for each function?

zqsz-xiv (commented):
Overall structural changes are fine with me, just had a few questions (and I haven't had the time to check all jobs' resources/skills for typos).

I got an error when trying to load this file:
[shanzhe_fru_p5_start.txt](https://github.com/user-attachments/files/18312069/shanzhe_fru_p5_start.txt)

image

Because of the size of this change, I'd like some further testing (whether automated or manual) before merging this in to make sure we don't break anybody's saved timelines. My personal test corpus covers BLM/PCT decently, but I don't really have any files for other jobs. If I have the time next week I'll see if I can transcribe some FRU logs for other jobs to test.

src/Game/Data/Jobs/BLM.ts
AkairyuGestalter: I'm not sure if these in particular are, since we tend not to actually use the resource name localization functions in the StatusPropsGenerators.

src/Game/Data/Jobs.ts
AkairyuGestalter: It's actually only used for defaulting the GCD cooldown in GameState's constructor. I'm not sure how useful that is since a given job can have both weapon skills and spells, but we can at least update that to infer based on role instead.

src/Game/Data/types.ts
AkairyuGestalter: Correct, further updated the comment to clarify.

src/Game/Data/types.ts
miyehn, line 1: I also took a look from high level and the overall direction looks so cool!!
* Are these interfaces used anywhere else other than with `ensureRecord`? If this is the case, can we name such interfaces slightly differently so they aren't confused with classes with the same names (Resource, Cooldown, etc.)? For example prepend them with capital I or something?
* Imo having a lot of data fields optional here is not great for readability because we then have to guess what `undefined` means for each of them.. so we might as well be clear and explicit even if it means a few more lines of code/change :0 At least just for the `ensureRecord` calls, we should be able to make the data fields required, then replace `ensureRecord<T>` with `ensureRecord<Partial<T>>`

AkairyuGestalter: Making the records allow `Partial<T>` would mean Typescript will never know definitively what properties the resultant object has, so I don't think that's a good idea.
Long term more of the properties should graduate to being required (like the action ID for example), but I left basically everything other than the English name that got cloned over from the string enums as optional to make the transition process able to be done in stages.
Very long term, my vision was that these interfaces could take the place of/become the basis for the existing skill/resource/cooldown types, at least for the basics like pre-speed cast times, max level cooldown times/charge counts, etc. The makeSkill calls could then get stuff like that from the Data object and the rest of the parameters should focus on what the skill does or handling traits and stuff. 

I can see about adjusting the names to prevent the name collision confusion though!

miyehn: Okay... as long as you are committed to the rest of migration then I think changes so far are fine to go in 👌

zqsz-xiv (approved):
My FRU P5 test corpus shows no changes between this branch and main for all our supported jobs, so I think we're good to merge this.

PR #95, opened 2024-12-24
Remove trait boilerplate and share test code

zqsz-xiv:
Moves all test code into `src/__test__` and helpers into `src/__test__/utils.ts`, and unifies setup into just three lines:
```ts

beforeEach(rotationTestSetup);
afterEach(rotationTestTeardown);
const testWithConfig = makeTestWithConfigFn(ShellJob.BLM);
```

Incidentally also removes some boilerplate for trait checks with `GameState.hasTraitUnlocked(TraitName)`, replacing `Traits.hasUnlocked(TraitName, level)` in many cases.


AkairyuGestalter (approved):
Overall lgtm and this didn't fall prey to my attempt that ended up still reporting test successes on cases that it really shouldn't have, so that's definitely something!

Also looks like this will mesh well with my drafted data updates and how I was going to simplify the Trait checking there as well, so that's a bonus.

PR #96, opened 2025-01-08
Initial GNB support for xivintheshell

namm100:
I added GNB support for xivintheshell!

Some things I have to fix:
- Warnings for certain actions
- No Mercy marker showing up on the two DoT tables.
- Multi-target splash damage

Thanks,

namm100 (xiv ign name: Sterling)


src/Game/Jobs/GNB.ts
```diff
@@ -0,0 +1,902 @@
...
+export class GNBState extends GameState {
+	constructor(config: GameConfig) {
+		super(config);
+        
+        // Enhanced Aurora adds an additional charge
+        const auroraStacks = this.hasTraitUnlocked(TraitName.EnhancedAurora) ? 2: 1;
+        [
+            new CoolDown(ResourceType.cd_Aurora, 60, auroraStacks, auroraStacks),
+        ].forEach((cd) => this.cooldowns.set(cd));
+        // const powderGaugeMax = this.hasTraitUnlocked(TraitName.CartridgeChargeII) ? 3: 2;
```
zqsz-xiv, line 122: did you remember to set this? it looks like even though at level 80 the gauge is visually capped at 2, you can hold up to 3 carts
[oops_3_carts.txt](https://github.com/user-attachments/files/18388188/oops_3_carts.txt)

src/Game/Jobs/GNB.ts
```diff
@@ -0,0 +1,902 @@
...
+
+	override jobSpecificAddDamageBuffCovers(node: ActionNode, skill: Skill<PlayerState>): void {
+		if (this.hasResourceAvailable(ResourceType.NoMercy)) {
+			node.addBuff(BuffType.NoMercy);
+		}
+		
+	}
+
+    // consume number of stacks and timer all in one
+    consumeTimer(resType: ResourceType, consumeStacks: number) {
```
zqsz-xiv, line 162: use `this.tryConsumeResource` instead; calling `Resource.consume(1)` when the resource is at 0 stacks causes a console warning

src/Game/Jobs/GNB.ts
```diff
@@ -0,0 +1,902 @@
...
+            if (skillName === SkillName.DemonSlice) {
+                this.resources.get(ResourceType.GNBAOEComboTracker).gain(1);
+                this.enqueueResourceDrop(ResourceType.GNBAOEComboTracker);
+            } 
+            return;
+        } 
+
+        // the other 3 combo types
+        var resType = ResourceType.GNBComboTracker;
+        var index = -1;
```
zqsz-xiv, line 187:
```suggestion
        let resType = ResourceType.GNBComboTracker;
        let index = -1;
```

src/Game/Jobs/GNB.ts
```diff
@@ -0,0 +1,902 @@
...
+                this.consumeTimer(resType, 2);
+            }
+        } else if (index === 2) {
+            this.consumeTimer(resType, 2);
+        }
+    }
+
+    // gain a cart
+    gainCartridge(carts: number) {
+        var maxCarts = this.hasTraitUnlocked(TraitName.CartridgeChargeII)? 3: 2;
```
zqsz-xiv, line 224:
```suggestion
        const maxCarts = this.hasTraitUnlocked(TraitName.CartridgeChargeII)? 3: 2;
```

src/Game/Jobs/GNB.ts
```diff
@@ -0,0 +1,902 @@
...
+            new CoolDown(ResourceType.cd_DoubleDown, this.config.adjustedSksGCD(60), 1, 1),
+        );
+        this.cooldowns.set(
+            new CoolDown(ResourceType.cd_GnashingFang, this.config.adjustedSksGCD(30), 1, 1),
+        );
+
+		// this.registerRecurringEvents();
+        super.registerRecurringEvents([ 
+            {
+                reportName: localizeResourceType(ResourceType.SonicBreakDoT),
```
zqsz-xiv, line 134:
```suggestion
```
removing this should remove the uptime %

src/Game/Jobs/GNB.ts
```diff
@@ -0,0 +1,902 @@
...
+                groupedDots: [
+                    {
+                        dotName: ResourceType.SonicBreakDoT,
+                        appliedBy: [SkillName.SonicBreak],
+                    },
+                ],
+            },
+            {
+                reportName: localizeResourceType(ResourceType.BowShockDoT),
```
zqsz-xiv, line 143:
```suggestion
```

zqsz-xiv (changes requested):
Nice work. Some additional high-level stuff:
- Remember to check build warnings, and run the formatter with `npm run format`
- Please add auto-replace for GF and its continuations so it's possible to save a GF+continuation pair as a preset, and load it to automatically become the correct ability. You should be able to do something similar to [what PCT does](https://github.com/xivintheshell/xivintheshell/blob/42ffa5887a75f423ec1cac23f4c74505363a306b/src/Game/Jobs/PCT.ts#L479-L480) for its filler combos that occur on the same button.
- Remember to add AoE falloff (just an additional optional `falloff` field to pass to the weaponskill/ability constructors). You can pretty much ctrl-F the job guide for "line" and "nearby" and paste in the % dropoff listed.
- Use `tryConsumeResource` instead of `state.consumeTimer` or `state.resources.get(...).consume(...)`; if you try to consume a resource currently at 0 it will raise a console warning, which happens for a lot of skills right now.
- IIRC the GF combo doesn't light up when you're doing a Reign Combo, but the converse is not true.
- Remember to add warnings on continuation drops + cart overcaps.
Also FYI #94, which is a pretty big change to how we reference skills + resource names, might get merged before this; I can help migrate to the new format if that gets in first.

zqsz-xiv:
Oh also, remember to set the animation lock of Trajectory to `MOVEMENT_SKILL_ANIMATION_LOCK`.

namm100:
I've implemented the majority of changes in the latest commit, sorry for the delay.
Issues remaining:
- No Mercy potency modifier tag not showing up on the dot tables. The calculations are correct however.
- I fiddled with src/Game/Common.ts a little more so for the Cartridge warning is a simple tweak.

Thanks!

zqsz-xiv (approved):
I went ahead and fixed the missing No Mercy tag in the DoT table + a few other random things. I tested a FRU P5 log and everything worked, so if you think it's ready to merge I'll hit the button.

For some reason GF combo + continuation presets don't auto-replace each other like PCT skills, but we can worry about that later. It might be something to do with the presence of a secondary cooldown.

namm100:
Sounds good to me! Let me know if any other issues/bugs I can fix arise or anything really. I'm fine with how it is so far.

PR #104, opened 2025-01-26
SMN in the Shell

zqsz-xiv:
This PR adds SMN to the Shell.
TODO:
- Colors are not final; I'll do a pass over them tomorrow
- Job gauge still includes debug information, and should be significantly simplified
- Test inputting the "cheat sheet" FRU timeline from #smn_encounter

System changes:
- `GameState.registerRecurringEvents` now takes an additional `petSkills` argument; this registers skills that summon pets as DoT effects to suppress a console assert in `DamageStatistics.ts` (this was being tripped for MCH's queen summon).
- The default DoT tick handler now checks if the buff is enabled (Slipstream can be toggled to simulate the boss stepping out of the puddle); if a tick occurs while the buff is disabled, then the tick count is incremented but the potency object for any skipped ticks is left unresolved.
- Skills that generate enmity now have a special `drawsAggro` flag in their declaration. This creates a 0-potency damage event on the timeline that is rendered if used while not in combat. This is useful if a SMN pulls with their demi-summon ability (which does not immediately do damage, but still draws aggro), or a tank pulls with Provoke or a non-damaging gap-closer like GNB's Trajectory. This will also be useful for healer DoTs, many of which deal 0 damage on initial application but still put the user in combat.

Implementation notes on pet logic (note to self: add this to wiki after merging):
- **Snapshots**: Pet damage (demi auto-attacks and egi/primal summons) are treated as DoT effects, much like MCH's Automaton Queen. Unlike standard DoTs, where tick damage inherits buffs from when it was first applied, pet abilities snapshot buffs at the _pet's_ "prepares" event rather than when the player presses the button, and this is reflected in this implementation.
- **Demis in-game** (Bahamut/Phoenix/Solar Bahamut): Per Hauffen (SMN mentor), "Enkindle" abilities (Akh Morn, Revelation, Exodus) that are executed by the pet are added to the pet's action queue, and execute after a varying amount of time. The amount of time between auto-attacks is generally near 3.163s, but varies for reasons that are unclear, and the rules surrounding pet auto timings and the exact enqueueing of the Enkindle action are unclear. Hauffen's hypothesis is that the auto-attack enqueued after an Enkindle has a shorter delay to compensate. (from his response to my question on [discord](https://discord.com/channels/277897135515762698/277968477233479680/1331449052344090724))
- **Demis in the shell**: I just implemented autos as a fixed timer on 3.163s intervals starting from the summon's confirmation, and the Enkindle ability executes immediately when pressed without affecting auto intervals. This is more or less consistent with the philosophy used by existing spreadsheets. We raise a warning if an Enkindle action is used within the last 2.5 seconds of the demi timer, as it becomes very likely to not be executed by the pet in time.
- **Primal and egi summons in-game** (Ifrit/Titan/Garuda): Using a "Summon [primal]" ability causes the primal to appear, do a big attack, and then leave; during this interval, other summons and Radiant Aegis cannot be used. This lockout typically lasts ~2.5 GCDs after the summon is used. At level 80 and below, summons may take longer to attack if the player is not in melee range when summoning, but overall take shorter time to leave than their level 90+ counterparts.
- **Primal and egi summons in the shell**: This implementation reflects in-game behavior fairly accurately, using rough numbers from a small sample size of timing data I logged myself. The durations shown in simulation more or less line up with my experience in-game.

[Sheet I used for summon + enkindle timings](https://docs.google.com/spreadsheets/d/1FSvf0n8Gbqb-95qbGwXtDLAo7ArxLvMDfwm_FAxvGA8/edit?gid=0#gid=0)


miyehn:
I didn't check the job's specifics but the system changes make sense 👌 

I think making a 0 potency damage marker to indicate draw aggro makes sense for now, or an alternative would be to make a different visual element type for draw aggro (a red or orange dot instead of a red triangle maybe). Or we can reconsider once healers need to use it (so the healing skills don't produce damage marks 🤔)

zqsz-xiv:
> an alternative would be to make a different visual element type for draw aggro (a red or orange dot instead of a red triangle maybe)

Yeah, I think that would be good.

I tested this against the recommended fight plan posted in #smn_encounter, and things check out (though some things in the plan don't quite line up with mechanics). I think this is OK to merge/deploy if there's no objections. [fru_smn_encounter_sheet_v1.txt](https://github.com/user-attachments/files/18555572/fru_smn_encounter_sheet_v1.txt)

PR #106, opened 2025-01-27
Test: automate loading fight plans from file and checking damage numbers

miyehn:
Looks like Github Actions isn't loading the newly added test files. I assume the failure will go away once it's in main?

Also @zqsz-xiv I recall you have a lot more transcribed fight plans. Would you be okay if I add them here as well?


miyehn:
Ohh good point about getting permission from the players first :0 I didn't think about they can take down their logs from fflogs but apparently not from here 😅 I do have explicit permission to use the two files included here for testing so I think they can stay. And let's not track the transcribed ones in git. This setup already solves my immediate problem which is to use them as sanity checks. I'll keep those files local for now.

Using a script to check between two commits and compare output in more granularity sounds great, it can be the next test suite upgrade when/if we get there >:D

zqsz-xiv (approved):

PR #107, opened 2025-01-28
SGE in the shell (initial pass)

AkairyuGestalter:
Core stuff added in addition to SGE's behavior:
- Added healing potency as a second potency kind. These show up on the timeline as green markers pointing up, instead of red markers pointing down.
- Added a special aggro marker as a sub-type of damage markers, now that the concept of aggro-only potencies exists. It's basically an orange damage marker.
- Updated the MP recovery calculation to handle Piety's effects for healers

Notes on SGE's implementation:
- AoE shields are currently assumed to have been applied to all applicable targets based on the Party Size resource. This is technically wrong with respect to game behavior, since single-target shields supersede AoE by virtue of the amount of shielding being done, but we don't currently have a good way to model this.
- Similarly, Eukrasian Diagnosis can only currently have a single application.
- Rather than make some silly counter, I have a longer term plan for modeling party members as targets for beneficial actions, but need more time to make that go.

This will at least get an initial version of SGE available for people to test with, and we'll wait on removing the Testing designation until I have had a chance to sort out the shield targeting sharp edges.


src/Components/Buffs.tsx
```diff
@@ -40,6 +40,12 @@ buffIcons.set(BuffType.ArcaneCircle, require("./Asset/Buffs/RPR/Arcane Circle.pn
 buffIcons.set(BuffType.RagingStrikes, require("./Asset/Buffs/BRD/Raging Strikes.png"));
 buffIcons.set(BuffType.Barrage, require("./Asset/Buffs/BRD/Barrage.png"));
 
+buffIcons.set(BuffType.Zoe, require("./Asset/Buffs/SGE/Zoe.png"));
```
zqsz-xiv, line 43: Did you intend for these to show up on healing skills that these affect? I couldn't see krasis/zoe on GCD heals when I tried using them.

src/Game/GameState.ts
```diff
@@ -503,6 +581,9 @@ export class GameState {
 		if (!this.isInCombat()) {
 			return 600;
 		}
+		if (this.job in HEALERS) {
```
zqsz-xiv, line 584: does this need to go before the `isInCombat` check?

src/Game/Jobs/DNC.ts
```diff
@@ -549,7 +548,6 @@ makeGCD_DNC("FOUNTAINFALL", 40, {
 		state.hasResourceAvailable("SILKEN_FLOW") || state.hasResourceAvailable("FLOURISHING_FLOW"),
 	onConfirm: (state) => {
 		if (state.hasResourceAvailable("SILKEN_FLOW")) {
-			state.tryConsumeResource("SILKEN_FLOW");
```
zqsz-xiv, line 552: accidental deletion?

src/Game/Jobs/SGE.ts
```diff
@@ -0,0 +1,1096 @@
...
+	override jobSpecificRegisterRecurringEvents() {
+		// Addersgall recurring timer
+		let recurringAddersgallGain = (rsc: Resource) => {
+			const addersgallInfo = getResourceInfo("SGE", "ADDERSGALL") as ResourceInfo;
+			// Not logging a warning since Rhizomata use can desync the timer from when it actually overcapped
+			if (!this.hasResourceAvailable("ADDERSGALL", addersgallInfo.maxValue)) {
+				this.resources.get("ADDERSGALL").gain(1);
+			}
+
+			this.resources.addResourceEvent({
```
zqsz-xiv, line 174: Just to be clear on the behavior here: the recurring timer is registered even if we're already capped, but the timer is then overridden when an ability that consumes addersgall is used?

src/Game/Jobs/SGE.ts
```diff
@@ -0,0 +1,1096 @@
...
+	const defaultRecastTime: ResourceCalculationFn<SGEState> = (state) =>
+		state.config.adjustedGCD(2.5);
+	const jobHealingPotencyModifiers: PotencyModifierFn<SGEState> = (state) => {
+		if (!params.healingPotency) {
+			return [];
+		}
+
+		const modifiers: PotencyModifier[] = [];
+
+		if (state.hasResourceAvailable("ZOE")) {
```
zqsz-xiv, line 342: Seems like this check for healing modifiers is being repeated a lot, can you put it into a function (or one function for healing GCDs + one for healing actions)?

src/Game/Jobs/SGE.ts
```diff
@@ -0,0 +1,1096 @@
...
+				kardiaPotency.modifiers.push(Modifiers.Krasis);
+			}
+			node.addHoTPotency(kardiaPotency, "KARDION");
+		},
+		(state, node) => {
+			if (!state.hasResourceAvailable("EUDAIMONIA")) {
+				return;
+			}
+
+			// Add Eudaimonia healing potency
```
zqsz-xiv, line 402: Does this mean each trigger of Eudaimonia snapshots healing buffs on the spell that triggers it, rather than when Philosophia itself is cast?

src/Game/Jobs/SGE.ts
```diff
@@ -0,0 +1,1096 @@
...
+			speedStat: "sps",
+		});
+		state.tryConsumeResource("EUKRASIA");
+	},
+	onApplication: (state, node) => {
+		state.applyDoT("EUKRASIAN_DOSIS_II", node);
+	},
+});
+
+makeSGESpell("EUKRASIAN_DOSIS_III", 72, {
```
zqsz-xiv, line 595:
```suggestion
makeSGESpell("EUKRASIAN_DOSIS_III", 82, {
```

src/Game/Jobs/SGE.ts
```diff
@@ -0,0 +1,1096 @@
...
+			newSkill: "EUKRASIAN_DYSKRASIA",
+		},
+	],
+	potency: 170,
+	falloff: 0,
+	applicationDelay: 0.76,
+	manaCost: 400,
+	castTime: 0,
+});
+makeSGESpell("EUKRASIAN_DYSKRASIA", 82, {
```
zqsz-xiv, line 725: needs falloff

src/Game/Jobs/SGE.ts
```diff
@@ -0,0 +1,1096 @@
...
+				(getResourceInfo("SGE", "HAIMA") as ResourceInfo).maxTimeout - 0.001, // Check just before the effect expires
+				() => state.handleHaimaExpiration("HAIMA", node),
+			),
+		);
+	},
+});
+makeSGEAbility("HAIMA_POP", 70, "cd_HAIMA_POP", {
+	startOnHotbar: false,
+	animationLock: FAKE_SKILL_ANIMATION_LOCK,
+	cooldown: 1,
```
zqsz-xiv, line 942: Is this cooldown accurate to in-game behavior? Right now if I spam haima/panhaima it forces a 1s delay between pops.

[poppoppoppop.txt](https://github.com/user-attachments/files/18627256/poppoppoppop.txt)

src/Game/Jobs/SGE.ts
```diff
@@ -0,0 +1,1096 @@
...
+	healingPotency: 600,
+	applicationDelay: 0.78,
+	validateAttempt: (state) => state.hasResourceAvailable("ADDERSGALL"),
+	onConfirm: (state) => state.consumeAddersgall(),
+});
+
+makeSGEAbility("IXOCHOLE", 52, "cd_IXOCHOLE", {
+	cooldown: 30,
+	healingPotency: 400,
+	applicationDelay: 0.78, // Assuming based on Druochole, the spreadsheet doesn't list it :(
```
zqsz-xiv, line 1044: I checked one of my random SGE logs and it had a 0.756s delay on applying to myself. I assume the sheet doesn't have a listed value because of propagation delay on party members.

zqsz-xiv (changes requested):
This is pretty big, thanks for the work.

A couple high-level questions:
1. Do you plan to add shielding potencies to the timeline mark?
2. Do you have a larger plan for using healing potencies/mitigation to manage HP, or anything else?

src/Game/Jobs/SGE.ts
AkairyuGestalter: Yeah, that's the expected behavior. This was yanked from BLM's Polyglot timer since they're logically similar.

src/Game/Jobs/SGE.ts
AkairyuGestalter: You're probably right, works for me.

src/Game/GameState.ts
AkairyuGestalter: Interestingly no. Out-of-combat MP recovery is a flat 600 MP per tick regardless of role. Piety only takes effect once you're in combat.

src/Game/Jobs/SGE.ts
AkairyuGestalter: That's correct. It functions the same as Kardia for all targets it's applied to, where the shapshot of the individual heal is based on the snapshot time of the action that triggered the heal, rather than the action that applied the status effect.

src/Components/Buffs.tsx
AkairyuGestalter: Right now they're only showing up on the timeline as skill buff covers/tooltips. I don't have a "healing statistics" table output to show them there the way we do for damage buffs at the moment.

src/Game/Jobs/SGE.ts
AkairyuGestalter: I'm not 100% sure on the exact number, but I'm pretty sure there is a hidden lockout before a second copy of the Haima shields can be consumed. If you have two separate damage sources hitting at nearly identical times, you'll still only lose one stack of the shield.

AkairyuGestalter:
Another thing that the follow-on "party member targeting" thing will need to deal with is single-target vs. aoe "healing action" buffs (ie. the distinction between Autophysis boosting heals for everyone vs. Krasis only affecting a single player, even for AoE heals/regen effects). Dealing with that at a core level is out of scope for this implementation.

zqsz-xiv:
Bumping my high-level questions (I think you might have missed them):

> 1. Do you plan to add shielding potencies to the timeline mark?
> 2. Do you have a larger plan for using healing potencies/mitigation to manage HP, or anything else?

Also, another question:
> Another thing that the follow-on "party member targeting" thing will need to deal with is single-target vs. aoe "healing action" buffs (ie. the distinction between Autophysis boosting heals for everyone vs. Krasis only affecting a single player, even for AoE heals/regen effects).

What's your intent for letting users accurately model AoE vs. single target healing? What would they be able to do that they can't under this healing implementation?

src/Components/TimelineCanvas.tsx
```diff
@@ -418,71 +418,118 @@ function drawWarningMarks(
...
+			case ElemType.HealingMark:
+				g_ctx.fillStyle = g_colors.timeline.healingMark;
+				time += localize({ en: "healing potency" });
+				break;
+			// If it's a damage mark, adjust color based on whether the boss is untargetable
+			default:
+				g_ctx.fillStyle = untargetable
+					? g_colors.timeline.untargetableDamageMark
+					: g_colors.timeline.damageMark;
+				time += localize({ en: "potency" });
```
zqsz-xiv, line 453: prefer "attack potency" or "damage potency" to make the distinction clearer

src/Game/Jobs/SGE.ts
```diff
@@ -0,0 +1,1098 @@
...
+				rscType: "ADDERSGALL",
+				name: "gain addersgall if not currently capped",
+				delay: addersgallInfo.maxTimeout,
+				fnOnRsc: recurringAddersgallGain,
+			});
+		};
+		recurringAddersgallGain(this.resources.get("ADDERSGALL"));
+	}
+
+	override jobSpecificAddHealingBuffCovers(node: ActionNode, _skill: Skill<PlayerState>): void {
```
zqsz-xiv, line 184: Zoe and philosophia steroids only affect GCDs, do you need to add a check for that here?

miyehn (commented):
(just cosmetic bits from messing with it-)

src/Game/Skills.ts
```diff
@@ -118,6 +118,13 @@ interface BaseSkill<T extends PlayerState> {
 	// check for undefined, not truthiness.
 	readonly falloff?: number;
 
+	readonly healingPotencyFn: ResourceCalculationFn<T>;
+	readonly jobHealingPotencyModifiers: PotencyModifierFn<T>;
+
+	// If defined, this button is treated as an AoE heal, and will apply that healing amount for the number
+	// of party members defined in the PARTY_SIZE resource, which defaults to 8
+	readonly aoeHeal?: boolean;
```
miyehn, line 126: We can make this just a boolean I think, instead of an optional boolean?

I'd always think twice before making anything optional, because it adds the additional value `undefined` to the type's domain, and it's not always clear what that `undefined` means --> reduces readability. It also basically requires everything that uses this value to be extra careful and specifically handle the `undefined` case...

The use of `undefined` in the above `falloff` is justified because there is a unique meaning attached to `undefined` and it's self-explanatory ("there is no falloff"), but in the case of `aoeHeal` it's not immediately clear what `undefined` means. Sure there is a comment above but 1) if it means the same thing as value `false`, we might as well just use `false`; 2) it's nicer if the code is self-explanatory xD

src/Components/ColorTheme.tsx
```diff
@@ -456,6 +491,8 @@ export let getCurrentThemeColors: () => ThemeColors = () => {
 				potCover: "#ff865c",
 				buffCover: "#9370db",
 				damageMark: "#ff0000",
+				healingMark: "#00ff00",
```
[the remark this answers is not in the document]
AkairyuGestalter: Just went with a full green channel since damage was full red :) I can't tell the difference with this so sure, works for me!

src/Components/TimelineCanvas.tsx
AkairyuGestalter: "damage potency" makes sense to me, since that's also how we're referring to it internally

src/Game/Jobs/SGE.ts
AkairyuGestalter: Argh, sure do. Some day I will deal with more tightly tying status effect resources to their buffs, but that day is not today.

src/Game/Skills.ts
AkairyuGestalter: That should be fine. I think I'd originally done the optional boolean to avoid breaking existing makeSkill functions, but that was before before fussing with the `MakeSkillParam` interface. Putting the null handling in the `make[Spell/Weaponskill/etc.]` calls instead of the `BaseSkill` interface does make downstream handling of the value easier.

AkairyuGestalter:
> Bumping my high-level questions (I think you might have missed them):
> 
> > 1. Do you plan to add shielding potencies to the timeline mark?
> > 2. Do you have a larger plan for using healing potencies/mitigation to manage HP, or anything else?
> 
> Also, another question:
> 
> > Another thing that the follow-on "party member targeting" thing will need to deal with is single-target vs. aoe "healing action" buffs (ie. the distinction between Autophysis boosting heals for everyone vs. Krasis only affecting a single player, even for AoE heals/regen effects).
> 
> What's your intent for letting users accurately model AoE vs. single target healing? What would they be able to do that they can't under this healing implementation?

I did miss this:
1. Not at the moment, but it might be useful. I didn't model the crit Diagnosis shield yet either since the exact shield values don't really matter at the moment, but it's a thing we could add.
2. At some point, adding a healing statistics table analogous to the damage statistics table should probably be on the roadmap. modeling HP as a resource probably not so much unless we were willing to also model boss damage which... ew.

The loose idea I was working with was updating status effect resource tracking for friendly effects to be a map of resources per [self, party member 1, party member 2, ..., party member n] and give some UI element for switching the current target for single-target effects. I haven't gone past the idea stage on it yet though.

zqsz-xiv:
> At some point, adding a healing statistics table analogous to the damage statistics table should probably be on the roadmap. modeling HP as a resource probably not so much unless we were willing to also model boss damage which... ew.

I'm going to push back on this a little. While I'm not a healer main, I'm not sure how many people would find a "healing potencies" table useful. A cooldown efficiency tracker like xivanalysis's "defensives" module might be nice, but just showing potencies + healing modifiers without the context of party health pools, incoming damage, and applied mitigations just doesn't seem to have much practical use. Tools like spreadsheet planners and ffxiv-fightline that make it easier to rearrange heals and cooldowns seem better suited for this purpose, especially considering how relatively unimportant GCD healing is in this game compared to mit cooldowns.

> The loose idea I was working with was updating status effect resource tracking for friendly effects to be a map of resources per [self, party member 1, party member 2, ..., party member n] and give some UI element for switching the current target for single-target effects. I haven't gone past the idea stage on it yet though.
> Edit: and the purpose behind the above would be for things like accurately tracking which/whether a target was affected by things like Krasis, banking a bunch of single-target Diagnosis shields to get multiple Addersting stacks off an AoE, and other stuff like that.

I do agree this could be nice to have more generally, especially if we want to more accurately display timeline-added party buffs affecting the current player, or targeted buffs like DNC's partner or spot mits for tanks/healers. That said, unless you think it's compelling, I don't think doing this just for the sake of generating healing statistics is all that useful.

Speaking from a super long-term perspective (probably 7.5/8.0 territory), I think eventually being able to model HP/boss damage and construct a mit timeline in xivintheshell is a pretty compelling story. If we go in that direction we'd have to model party AoE heal potencies and buffs either way, but absent a plan for that, I'm currently not convinced about its utility as a standalone feature.

src/Game/Skills.ts
zqsz-xiv: We should also make this adjustment later for the `requiresCombat` flag, since it's also of type `undefined | boolean`. Even for `falloff` (which like miyehn said has a clear meaning for undefined) I was debating using a default value of 1.0 to represent skills that only damage a single target, but in the end I figured that would be just as confusing as checking for undefined.

miyehn:
By running shanzhe's test corpus locally I just found that `the_jz_rdm_fru_p5.txt` can't be loaded on this branch :0

zqsz-xiv:
> By running shanzhe's test corpus locally I just found that the_jz_rdm_fru_p5.txt can't be loaded on this branch :0

It looks like this is a bug on main, where the `requiresCombat` flag isn't being passed properly within `makeResourceAbility`, which this PR fixes by combining the parameters. RDM is the only job affected because it calls `makeResourceAbility` for Manafication, while all other jobs with skills that require combat use `makeAbility` directly.

AkairyuGestalter:
> > By running shanzhe's test corpus locally I just found that the_jz_rdm_fru_p5.txt can't be loaded on this branch :0
> 
> It looks like this is a bug on main, where the `requiresCombat` flag isn't being passed properly within `makeResourceAbility`, which this PR fixes by combining the parameters. RDM is the only job affected because it calls `makeResourceAbility` for Manafication, while all other jobs with skills that require combat use `makeAbility` directly.

Not sure if there's something to still fix here, or if this was a "should have broken on main before, but wasn't because oops" situation, but if it's the former let me know!

The thing that seems to have gotten lost along the way on the discussion around the party target selection thing was the original issue of this implementation not being able to handle properly stacking/popping multiple EDiag shields for Addersting and also not properly modeling a given target only being able to be affected by either EDiag or EProg (or Galvanize but lets not get into that right now). I think that chatter might've happened over on the Discord dev thread so it's easy to lose. That plus the issues around Krasis being a single-target heal buff that shouldn't affect all AoE heal potencies from a given action (AST's current Arrow card is going to have the same problem) are the original impetus behind adding a "select party member target" system, since I also didn't want a band-aid "pretend EDiag stacks" or "grant arbitrary number of Addersting" button solution to the shield pop side of the problem. Setting aside the hypothetical healing potency statistics output which wasn't really something I had active plans to do, I still think there'll be value for this for accurately modeling SGE (and AST) behavior. It just doesn't belong on _this_ PR.

zqsz-xiv:
> Not sure if there's something to still fix here, or if this was a "should have broken on main before, but wasn't because oops" situation, but if it's the former let me know!

It's the latter (should have broken on main before).

> Setting aside the hypothetical healing potency statistics output which wasn't really something I had active plans to do, I still think there'll be value for this for accurately modeling SGE (and AST) behavior. It just doesn't belong on this PR.

Makes sense, I just wanted to make sure you didn't over-engineer something in the future if there wasn't a concrete use for it.

AkairyuGestalter:
> Makes sense, I just wanted to make sure you didn't over-engineer something in the future if there wasn't a concrete use for it.

Yeah, definitely had a purpose in mind, just wasn't doing a good job of articulating it :)

miyehn:
👀 we still waiting for changes here (other than resolving conflicts)?

zqsz-xiv:
No further comments on my end.

AkairyuGestalter:
I'll get that conflict resolved and a changelog entry made tonight. I got heckin' distracted the last couple weekends messing around with Ocarina of Time randomizers...

zqsz-xiv (approved):
Looks good, just needs the non-combat `drawsAggro` flag for e. dosis and e. dyskrasia.

zqsz-xiv:
Oh, can you also add this changelog entry (since it's fixed incidentally)?

- [RDM] Fixed a bug where Manafication could be used before combat

miyehn:
Oh and you should add SGE to the list of jobs you added/maintain as well 👀 (SGE is 贤者 in Chinese)

PR #108, opened 2025-01-29
Initial DRG implementation

namm100:
Hello!

Here is an initial pass for DRG in the shell. 3 known issues to be worked on:
- Mousing over a skill doesn't show the buff image icons (GNB also has this issue).
- DoT table doesn't include buff labels and is adding combo/positionals to the tooltip but calculating everything correctly.
- Starcross Ready needs to be consumed once Life of the Dragon timer is finished.

Thanks!

namm100 (Sterling)


src/Components/Jobs/DRG.tsx
```diff
@@ -0,0 +1,87 @@
...
+				kind: "bar",
+				name: localize({ en: "Life of the Dragon", zh: "LT" }),
+				color: colors.rdm.manaStack,
+				progress: lifeTimer ? lifeTimer / 20 : 0, // TODO EXACT VALUE
+				valueString: lifeTimer?.toFixed(3) ?? "N/A",
+			} as ResourceBarProps,
+
+			{
+				kind: "counter",
+				name: localize({ en: "Firstmind's Focus", zh: "FF" }),
```
zqsz-xiv, line 78:
```suggestion
				name: localize({ en: "Firstmind's Focus" }),
```

src/Components/Jobs/DRG.tsx
```diff
@@ -0,0 +1,87 @@
...
+				kind: "bar",
+				name: localize({ en: "Combo Timer", zh: "CT " }),
+				color: colors.rdm.manaStack,
+				progress: comboTimer ? comboTimer / 30 : 0,
+				valueString: comboTimer?.toFixed(3) ?? "N/A",
+			} as ResourceBarProps,
+
+			{
+				kind: "bar",
+				name: localize({ en: "Life of the Dragon", zh: "LT" }),
```
zqsz-xiv, line 70:
```suggestion
				name: localize({ en: "Life of the Dragon" }),
```

src/Components/Jobs/DRG.tsx
```diff
@@ -0,0 +1,87 @@
...
+			: heavensCombo.available(1)
+				? heavensCombo.pendingChange?.timeTillEvent
+				: aoeCombo.available(1)
+					? aoeCombo.pendingChange?.timeTillEvent
+					: undefined;
+
+		const infos: ResourceDisplayProps[] = [
+			{
+				kind: "bar",
+				name: localize({ en: "Combo Timer", zh: "CT " }),
```
zqsz-xiv, line 62:
```suggestion
				name: localize({ en: "Combo Timer" }),
```

src/Components/Jobs/DRG.tsx
```diff
@@ -0,0 +1,87 @@
...
+				color: colors.rdm.manaStack,
+				progress: comboTimer ? comboTimer / 30 : 0,
+				valueString: comboTimer?.toFixed(3) ?? "N/A",
+			} as ResourceBarProps,
+
+			{
+				kind: "bar",
+				name: localize({ en: "Life of the Dragon", zh: "LT" }),
+				color: colors.rdm.manaStack,
+				progress: lifeTimer ? lifeTimer / 20 : 0, // TODO EXACT VALUE
```
zqsz-xiv, line 72: You should flip this to (20 - lifeTimer) so it fills left to right. Up to you if you want this behavior for the combo timer as well.

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+				}
+			}
+			this.resources.addResourceEvent({
+				rscType: "STARCROSS_READY",
+				name: "drop starcross outside life of the dragon",
+				fnOnRsc: dropStarcrossOutsideLife,
+				delay: 1,
+			});
+		};
+		dropStarcrossOutsideLife(this.resources.get("STARCROSS_READY"));
```
zqsz-xiv, line 144: Instead of making this a recurring event, when you call `this.addEvent` to begin the life window, you can add a second parallel event that calls `this.tryConsumeResource("STARCROSS_READY")` when the life window ends.

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+		}
+
+		let combo_value = -1;
+		// chaos combo
+		if (CHAOS_COMBO_MAP.has(skillName)) {
+			this.tryConsumeResource("DRG_HEAVENS_COMBO_TRACKER", true);
+			resType = "DRG_CHAOS_COMBO_TRACKER";
+			combo_value = CHAOS_COMBO_MAP.get(skillName)
+				? Number(CHAOS_COMBO_MAP.get(skillName))
+				: -1;
```
zqsz-xiv, line 220:
```suggestion
			combo_value = CHAOS_COMBO_MAP.get(skillName) ?? -1;
```

Using the `Number` constructor is a bit of an anti-pattern, since it tends to hide possible logic errors.

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+		}
+		// heavens combo
+		else if (HEAVENS_COMBO_MAP.has(skillName)) {
+			this.tryConsumeResource("DRG_CHAOS_COMBO_TRACKER", true);
+			resType = "DRG_HEAVENS_COMBO_TRACKER";
+			combo_value = HEAVENS_COMBO_MAP.get(skillName)
+				? Number(HEAVENS_COMBO_MAP.get(skillName))
+				: -1;
```
zqsz-xiv, line 228:
```suggestion
			combo_value = HEAVENS_COMBO_MAP.get(skillName) ?? -1;
```

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+
+		if (this.resources.get(resType).availableAmount() === combo_value) {
+			this.resources.get(resType).gain(1);
+			this.enqueueResourceDrop(resType);
+		} else {
+			this.tryConsumeResource(resType, true);
+		}
+	}
+
+	// Return true if the active combo buff is up, or meikyo is active.
```
zqsz-xiv, line 245:
```suggestion
	// Return true if the active combo buff is up.
```

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+					params.combo.resourceValue
+			) {
+				mods.push(
+					makeComboModifier(
+						getBasePotency(state, params.combo.potency) -
+							getBasePotency(state, params.potency),
+					),
+				);
+			}
+            */
```
zqsz-xiv, line 350:
```suggestion
```

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+		resourceValue: 1,
+	},
+	onConfirm: (state) => {
+		if (state.resources.get("DRG_CHAOS_COMBO_TRACKER").availableAmount() === 1) {
+			state.refreshBuff("POWER_SURGE", 0);
+		}
+	},
+	onApplication: (state) => {
+		// DRG is weird with power surge
+		// when disembowel is applied combo tracker should have already progressed
```
zqsz-xiv, line 560: For situations like this where you need to snapshot the combo state at confirmation rather than application, you should call `state.refreshBuff` with the application delay in the `onConfirm` function, like what SAM does for fugetsu/fuka. Right now you're refreshing the buff immediately both on confirm and application, which isn't quite right.

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+			node,
+			dotName: "CHAOS_THRUST_DOT",
+			skillName: "CHAOS_THRUST",
+			tickPotency,
+			speedStat: "sks",
+			modifiers,
+		});
+	},
+	onApplication: (state, node) => {
+		state.applyDoT("CHAOS_THRUST_DOT", node);
```
zqsz-xiv, line 630: The tooltip seems to imply that the DoT only applies if the ability is combod. Like with buff refreshes, you should instead call `applyDoT` from within `onConfirm` on a delay after checking the combo state there.

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+			node,
+			dotName: "CHAOTIC_SPRING_DOT",
+			skillName: "CHAOTIC_SPRING",
+			tickPotency,
+			speedStat: "sks",
+			modifiers,
+		});
+	},
+	onApplication: (state, node) => {
+		state.applyDoT("CHAOTIC_SPRING_DOT", node);
```
zqsz-xiv, line 673: same comment about snapshot timing with chaos thrust

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+	combo: {
+		potency: 120,
+		resource: "DRG_AOE_COMBO_TRACKER",
+		resourceValue: 1,
+	},
+	falloff: 0,
+	highlightIf: (state) => state.resources.get("DRG_AOE_COMBO_TRACKER").availableAmount() === 1,
+	onConfirm: (state) => {
+		if (state.resources.get("DRG_AOE_COMBO_TRACKER").availableAmount() === 1) {
+			state.refreshBuff("POWER_SURGE", 0);
```
zqsz-xiv, line 864: should this be on a delay?

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+		resourceValue: 2,
+	},
+	falloff: 0,
+	highlightIf: (state) => state.resources.get("DRG_AOE_COMBO_TRACKER").availableAmount() === 2,
+	onConfirm: (state) => {
+		if (
+			state.hasTraitUnlocked("ENHANCED_COERTHAN_TORMENT") &&
+			state.resources.get("DRG_AOE_COMBO_TRACKER").availableAmount() === 2
+		) {
+			state.refreshBuff("DRACONIAN_FIRE", 0);
```
zqsz-xiv, line 883: should this be on a delay?

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+			resource: DRGResourceKey;
+			resourceValue: number;
+		};
+		positional?: {
+			potency: number | Array<[TraitKey, number]>;
+			comboPotency: number | Array<[TraitKey, number]>;
+			location: "flank" | "rear";
+		};
+		jobPotencyModifiers?: PotencyModifierFn<DRGState>;
+		applicationDelay?: number;
```
zqsz-xiv, line 301: You should make `applicationDelay` mandatory to ensure you don't forget to set it for a skill.

```suggestion
		applicationDelay: number;
```

src/Game/Jobs/DRG.ts
```diff
@@ -0,0 +1,1069 @@
...
+	params: {
+		autoUpgrade?: SkillAutoReplace;
+		autoDowngrade?: SkillAutoReplace;
+		requiresCombat?: boolean;
+		potency?: number | Array<[TraitKey, number]>;
+		falloff?: number;
+		replaceIf?: ConditionalSkillReplace<DRGState>[];
+		highlightIf?: StatePredicate<DRGState>;
+		startOnHotbar?: boolean;
+		applicationDelay?: number;
```
zqsz-xiv, line 422:
```suggestion
		applicationDelay: number;
```

zqsz-xiv (changes requested):
Thanks for the PR. In general, remember to specify an application delay for all skills, and if you need to snapshot state (like combo progress) at confirmation time, then you should make a call to `state.addEvent` (or your `refreshBuff` helper) from within the `onConfirm` function.

A couple more things:
- It looks like Enhanced Piercing Talon and Power Surge are trying (but failing) to produce a buff image on the timeline when hovered. For any buffs you want to show in this way, remember to register them in `src/Components/Buff.tsx`
image (I think I forgot to tell you to do this for GNB).
- If you have the time, please transcribe (or create) a timeline of a few minutes from a log (I have a FRU P5 timeline for every job that I use to test things). This will help catch any outlying bugs or edge cases you may have missed.

src/Game/Jobs/DRG.ts
namm100: Screenshot 2025-02-02 at 9 15 52 AM
This one looks like its on the `onConfirm` window.

src/Components/Jobs/DRG.tsx
namm100: Screenshot 2025-02-02 at 9 52 41 AM

In game it does fill right to left but honestly I'm unsure which one I prefer haha.

namm100: I'm pretty sure in game it fills right to left but for practical purposes, I'm honestly unsure which one I prefer.

src/Game/Jobs/DRG.ts
namm100: Screenshot 2025-02-02 at 10 10 18 AM

Here's Coerthan Torment

src/Game/Jobs/DRG.ts
namm100: This is an interesting point - I was discussing this with q'ras but it seems that power surge is applied in a weird way, I forgot to copy-paste implement it for Sonic Thrust.

To make it exactly like in game, power surge is applied at the `onConfirm` window and the timer doesn't queue until the `onApplication` window. For now I just refresh the buff on both windows.

namm100:
Still working on buff icons/stuff, but the small changes are pushed.

src/Components/Jobs/DRG.tsx
zqsz-xiv: Oh, it looks like you're right. Up to you then.

src/Game/Jobs/DRG.ts
zqsz-xiv: It might be one of those cases where the actual duration of the buff is longer than the applied length. It looks like the DRG planner sheet in the balance has a duration of 30.62s for Power Surge, and 20.62 for Lance Charge + Litany.

https://docs.google.com/spreadsheets/d/19YcFZnVJ707CtnJI-waEFJkqf3NMg-3ooWa2hhg4KiU/edit?gid=1199497531#gid=1199497531

zqsz-xiv: Actually never mind, those might just be adding animation lock to account for application delay. To test it for sure you should run a log and let the buff expire, and check the actual duration between the "apply buff"
 and "remove buff" events.

namm100:
Final Issue(s) I'm aware of:
- Dot tick potency incorrectly modified by additive increases in the tooltip but the computation is correct.
- Need to write a test log/test rotation.

namm100:
Added a test rotation from based on https://www.fflogs.com/reports/V6xaMgjHyGFb9tcP?fight=9&type=casts&view=events&source=31

I wonder if dive ready is a bit more than 15 sec...

zqsz-xiv (commented):
Mostly looks good, just some minor things.

As for test logs: we've decided to keep timelines that are transcribed directly from fflogs out of github, so as to avoid a potential scenario where somebody wants to delist the log and aren't OK with us reproducing it here. As such, could you remove the transcribed log from `__test__/Asset`? I've downloaded it for manual testing later, and the fact that you were able to successfully transcribe a log is also a good test in itself.

If you want to add tests for specific scenarios (like double mirage dive or power surge lasting for 2 combos), then you should put together a short timeline with that portion, and make a test in a new `DrgRotations.test.ts` file like [this one](https://github.com/xivintheshell/xivintheshell/blob/52218c30e18d5ebab221172c9f9d72fe1ddb4939/src/__test__/BlmRotations.test.ts#L596).

zqsz-xiv (approved):

PR #111, opened 2025-02-22
Boilerplate {Do you need it?}

AkairyuGestalter:
Just stubbing out the remaining `StatusPropGenerator` and `GameState` classes for the jobs that haven't been implemented yet. Nothing fancy, just less work for any other new developers that want to contribute.


miyehn (approved):
doesn't hurt 🎉🚀

PR #113, opened 2025-02-27
New script setup to help coordinate localization effort

miyehn:
This PR is just for the script's setup which is a bunch of miscellaneous changes. The actual script is not important here
- put back `ts-node` dependency so we can have scripts with ts code
- inlined `hasTraitUnlocked` utility function, to get rid of a circular dependency. Seems like it hasn't been an issue because webpack is pretty smart at resolving dependencies but `ts-node` just blindly imports everything listed at each file's top
- fixed a typo in BRD's skill name

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,7 @@
         "react-tabs": "^4.0.1",
         "react-tooltip": "^5.8.3",
         "seedrandom": "^3.0.5",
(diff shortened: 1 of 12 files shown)


miyehn:
Yuyuka already copy-pasted lines of code into the excel sheet so I guess we are good for now

PR #115, opened 2025-03-14
Move TESTING/OUTDATED warning copy into Support Interface

AkairyuGestalter:
Mark fully-supported jobs as outdated due to patch 7.2 changes to warn users about instability until we can update them all.

Notes: 
- We probably want a CN translation for the OUTDATED warning content before we push this out.
- We probably don't want to push this out before the servers go down for the patch.
- Submitting as a draft for right now because of these.

I also have no idea what we're going to do about the CN/KR regions lagging behind Global given the scope of at least the BLM changes...

### DIFF
--- a/src/Components/PlaybackControl.tsx
+++ b/src/Components/PlaybackControl.tsx
@@ -222,28 +222,17 @@ export function ConfigSummary(props: { job: ShellJob; dirty: boolean }) {
 	});
 	let warningColor = getCurrentThemeColors().warning;
 	let tryGetImplementationWarning = (impl: ImplementationKey, warningColor: string) => {
-		if (impl === "TESTING") {
-			return <p
-				style={{
-					color: warningColor,
-				}}
-			>
-				{localize({
-					en: "WARNING: This job was recently added to XIV in the Shell and is still being tested. There may be bugs or changes in the near future, so make sure to frequently export and save timelines for this job to make sure you don't lose your work.",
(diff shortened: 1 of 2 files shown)


miyehn:
Proposed OUTDATED blurb translation:
警告：此职业最近经历的技改还未被完整实现到排轴器中。当前排轴器的此职业可能有bug，或在近期经历更新，所以暂时不要太依赖txt，记得勤在别处保存排轴进度。

I would probably also rephrase the English blurb to ask players not to rely on txt files until the job is fully updated, because old txt files may no longer load correctly 🤔 

Historically at least BLM in the Shell just updated together with the global regions, and assumed the CN/KR players would get an account in a global server if they cared about theorycrafting this much 😂

AkairyuGestalter:
> 警告：此职业最近经历的技改还未被完整实现到排轴器中。当前排轴器的此职业可能有bug，或在近期经历更新，所以暂时不要太依赖txt，记得勤在别处保存排轴进度。

Added the suggested zh translation and reworded the English warning. If you need or want a revision to the zh copy after reviewing my en change, let me know :)

AkairyuGestalter:
Reverted the `OUTDATED` designation for all jobs but BLM, since the only changes were to potencies and pre-7.2 timelines will still work post 7.2. BLM needs work though, so leaving that as `OUTDATED` until we update it.

AkairyuGestalter:
Since shanze's going to have BLM updated shortly, PR is now just moving the location of the warning text and adding the ZH translation for the OUTDATED implementation type.

PR #121, opened 2025-03-20
Add GNB action IDs

AkairyuGestalter:
Title.

Also a fix to the name of GNB's ranged attack "Lightning Shot" vs. the typoed "Lightning Shock". Didn't change the key of the action to keep from breaking existing plans. Should probably fix if/when we have larger patch changes that would break pre-existing plans anyways. Definitely with 8.0...


zqsz-xiv (commented):
> Also a fix to the name of GNB's ranged attack "Lightning Shot" vs. the typoed "Lightning Shock". Didn't change the key of the action to keep from breaking existing plans. Should probably fix if/when we have larger patch changes that would break pre-existing plans anyways. Definitely with 8.0...

You could add a clause to `getNormalizedSkillName` in `Skills.ts` to handle this. Also, since serialization stores the value (not the action key) I think this would break existing timelines.

zqsz-xiv:
@AkairyuGestalter I fixed the naming issue and added a thing to make sure old timelines don't get broken. Can you rebase and update this branch?

zqsz-xiv (approved):

PR #124, opened 2025-03-25
7.2 you can (not) drop enochian

zqsz-xiv:


misterperson (commented):
Signing off on /src/Game/Jobs/WAR.ts changes.

zqsz-xiv:
Got signoff from other devs on dc, merging and deploying now.

Issue #126, opened 2025-04-13
[RPR] Add gauge gain from target dying with DD


zqsz-xiv:
From the user survey:

> Reaper gains Gauge from things dying under DD, which is not accounted for in the planner

We can add a fake gauge gain button similar to "pop" actions on other jobs.

PR #129, opened 2025-04-17
[RPR] Add soul gauge gain button

zqsz-xiv:
#126


Junie-774:
Looks good to me! I like the icon

PR #137, opened 2025-04-29
Resource and CoolDown refactors

miyehn:
Made Resource and CoolDown classes more separate by pulling some some functions out of the base abstract class
Got rid of recast time scaling for CDs, and switch to just store number of stacks available + time till next stack

As a result, there are slight changes to some jobs' implementations
* BLM: Ley Lines is now handled in `Game/Jobs/BLM.ts`
* PCT: Tempera Coat and Tempera Grassa - hopefully I changed them right?
* Others: remove redundant function call parameter(s)

Site behavior is the same as before according to some basic manual testing and shanzhe's test corpus. Resource overrides for CDs also still work.

And I slipped in a few translations

### DIFF
--- a/src/Game/Resources.ts
+++ b/src/Game/Resources.ts
@@ -186,8 +186,10 @@ export class CoolDown extends ResourceOrCooldown {
 			if (this.#timeTillNextStackAvailable < Debug.epsilon) {
 				super.gain(1);
 				this.#currentRecast = this.#defaultRecast;
-				if (this.stacksAvailable() <= this.maxStacks()) {
+				if (this.stacksAvailable() < this.maxStacks()) {
 					this.#timeTillNextStackAvailable = this.#defaultRecast;
+				} else {
+					this.#timeTillNextStackAvailable = 0;
 				}
 			}
 			deltaTime -= forThisStack;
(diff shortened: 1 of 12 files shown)


src/Game/Resources.ts
```diff
@@ -131,75 +146,61 @@ export class OverTimeBuff extends Resource {
...
-			this.#reCaptureRecastTimeScale(game);
-			// and part after stack gain
-			this.gain((deltaTime - scaledTimeTillNextStack) / this.#recastTimeScale);
-		} else {
-			this.gain(deltaTime / this.#recastTimeScale);
+		this.#currentRecast = recast;
+		this.useStack();
+	}
+	restore(deltaTime: number) {
+		while (deltaTime > 0 && super.availableAmount() < this.maxStacks()) {
```
zqsz-xiv, line 183: why does this need a `while` loop? is it possible to express this with arithmetic instead?

zqsz-xiv (approved):
Everything looks good, just one question.

src/Game/Resources.ts
miyehn: It's probably possible, I just find while loop conceptually simple, because every iteration only advances enough time without skipping anything interesting. It's like `GameState.tick()`. And it's only going to run that many iterations because `thismaxStacks()` can't be too large

PR #140, opened 2025-05-02
framework updates

zqsz-xiv:
Not ready for review, just putting this up to track tasks and things to test. The upgrades here were compelled by an attempt to install base-ui: https://base-ui.com/react/overview/quick-start that was stymied because we have outdated dependencies constrained by create-react-app.

The more I use nextjs the more I'm convinced that it's a bad fit as a create-react-app successor, but we don't have a whole lot of choice... maybe react-router?

Things to test:
- missing skill icon rendered in hotbar, presets, and canvas if a path is specified incorrectly
- tests work
- on-the-fly dark mode switching works
- page reset command URLs work
- 404 routes work

probably can remove jquery while we're at it, since it only exists to do XHR


miyehn:
Sounds painful.... 🙏 I wish I could suggest better ways to profile and optimize but web & browser performance is also a mystery to me :/ and it doesn't look fun/easy to go arbitrarily low-level to figure out why..
The only thing that came to mind was I remember being suggested to use actual css & element (class) selectors over style objects in js which mostly get turned into inline element style, because the former is supposed to be faster. But that was at least a few years ago and I never compared the difference myself (becauce it was already performant enough). So not sure how relevant here...

zqsz-xiv:
That's a good point, though style rendering is probably not the bottleneck here.

I did some more testing, and it seems like running the production build (`npx react-router build && vite preview --open`) doesn't have any of these perf issues, and the page is as snappy as ever. Looking at the browser profiler in the dev build, it seems like there are a ton of calls to some `createTask` function that seems to be part of the react-router dev runtime. There might be some configuration option to disable it, but the documentation isn't great and I haven't found it yet.

I'm still a little hesitant to push this change because it will make it hard for us to identify any performance issues in the future, but at the very least we can proceed with the dependency upgrade without doing anything overly dramatic.

zqsz-xiv:
I found the time to dig deeper into figuring out how to improve performance, and ended up at a point where I'm satisfied enough to merge. After removing react-router and doing some other configuration changes, most of the dev build is back to being fairly responsive. Skill input is still a little sluggish, but it's more tolerable now.

Remaining stuff to test before merge:
- I broke image export (it seems to be spitting out an un-renderable HTML file), so I need to go back and fix that
- Test corpus
- etro/xivgear import
- URL commands (these have been switched to search parameters like `xivintheshell.com?command` instead of a URL path like `xivintheshell.com/command`)

For posterity, here's a breakdown of the process I went through:

## What did I do, and why?

Though I couldn't find any information about the specific "createTask" function I was observing in the profiler, the discrepancy between dev and production builds was enough to give me some leads. Vite has a short [page on performance](https://vite.dev/guide/performance.html#performance) that was very helpful. In the end, the steps I took basically aligned with the bullet points on that page.

### "Audit Configured Vite Plugins"

Since it seemed like the performance bottlenecks were coming from outside our own code, I turned to [`vite-plugin-inspect`](https://github.com/antfu-collective/vite-plugin-inspect) to help identify what react-router and vite were doing. The results of plugin actions performed on a page refresh looked something like this:

plugins_with_react_router

My understanding is that the `esbuild` task is responsible for a significant chunk of transpilation work, and apparently in dev builds it's run on some amount of code every time a component renders. Vite's perf tips also note this:

> When using @vitejs/plugin-react, avoid configuring the Babel options, so it skips the transformation during build (only esbuild will be used).

Note the presence of the `react-refresh-babel` plugin; while it doesn't dominate the runtime, it is significant all the same. It seems like the relationship between react-router's vite plugin and `@vitejs/plugin-react` is rather complicated--based on this [GH thread](https://github.com/remix-run/react-router/issues/12352), react-router vendors their own copy of the vendor to better support some live module reload features, but this has the side effect of bringing in Babel configuration in a way that the end user cannot touch. Based on the thread, it seems like this is an area that react-router's maintainers are hoping to improve, so in the near future it may be possible for us to return to react-router. In the meantime, however, I decided to try removing react-router and replacing its functionalities with native browser APIs. After doing so, the plugin profile looked like this:

plugins_no_router_normal

This was a respectable improvement, but could still be improved. This is where the rest of the perf tips came in...

### "Use Lesser or Native Tooling"

Swapping in `@vitejs/react-plugin-swc` in place of `@vitejs/react-plugin` enables the faster SWC transpiler, and this in turn gave another decent perf gain for free.

plugins_no_router_swc

This is part of what motivated me to remove react-router; the react-router vite plugin seemingly had no way to enable SWC instead. I also tried the experimental [rolldown](https://vite.dev/guide/rolldown.html) bundler, but when used together with react-router this caused `vite dev` to panic, and in the absence of react-router didn't make an appreciable difference when compared to SWC alone. This could be a promising option when it becomes more mature though.

All this led to one final performance optimization...

### "Reduce Resolve Operations"

From the vite perf guide:
```
When you try to import ./Component.jsx with import './Component', Vite will run these steps to resolve it:

1. Check if ./Component exists, no.
2. Check if ./Component.mjs exists, no.
3. Check if ./Component.js exists, no.
4. Check if ./Component.mts exists, no.
5. Check if ./Component.ts exists, no.
6. Check if ./Component.jsx exists, yes!

As shown, a total of 6 filesystem checks is required to resolve an import path. The more implicit imports you have, the more time it adds up to resolve the paths.
```

Since almost all our files are `.tsx` or `.ts`, we stand to benefit significantly from a more intelligent file resolution order, which vite thankfully lets us configure. Now, the profile looks like this:
plugins_no_router_swc_resolves

This is much better than before. Oddly enough, changing the resolution order did not make an appreciable difference when using the react-router plugin. It's possible that the extension resolution option did not reach deep enough to affect the babel plugin that react-router uses.

In conclusion: we don't need to rewrite the world quite yet. With or without react-router, the production build runs fine, but removing it and fiddling with these configuration options seems to have made the dev build much faster.

zqsz-xiv:
Bugs are fixed, and tests are passing. I also replaced `react-tooltip` with the BaseUI tooltip component (since I'm planning to bring in the Dialog component from BaseUI for announcements anyway), and shaved 100 kB off the final bundle.

I'll work on adding the updated changelog/announcements format in a separate branch so this PR doesn't get too bloated, but if I finish it up within the next couple days I'll probably toss it in this PR as well. I'd like to have the new changelog system when this refactor deployed so users have something to blame for any bugs, and so I have an excuse to make an announcement for the existence of the new changelog format.

PR #145, opened 2025-05-14
More prominent changelog

zqsz-xiv:
(branched off #140)

This PR moves the changelog tab out of the intro section, and makes it more prominent. The new changelog opens in a dialog box displaying the 5 most recent entries, or all entries since the user's last view of the changelog, with older entries in an expandable. Additionally, this PR adds boilerplate for localized changelog entries.

Example gif (shows basic behavior, page refresh, and localizable entries):
full_changelog_demo

Internal changes:
- Replaced `react-tooltip` by `@base-ui-components`'s tooltip component. This reduces the overall bundle size, and the JS bundle with both `@base-ui-components/react/tooltip` and `@base-ui-components/react/dialog` is about 100 kB smaller before gzip. Some additional boilerplate is necessary, but component arrangement is mostly easier to reason about.
- Added some additional CSS classes.
- Added the `isFirstVisit` variable, determined by whether localstorage is empty on module load. I assume the declaration of the variable always runs before any `setCachedValue` operations, unless there's some weird hoisting shenanigans going on.

Behavior flowchart (can test "first time visitor" flow by opening in private/incognito):
- If user has `changelogLastRead` in localstorage
  - If `changelogLastRead` matches the date in the most recent entry:
    - If `changelogLastReadChangeCount` matches the number of elements in the most recent entry's `changes` field, then the Changelog is not highlighted.
    - Otherwise, the Changelog button is stylized bold + warning color + exclamation points.
  - If `changelogLastRead` does NOT match the date in the most recent entry, the Changelog is not highlighted.
- If user does NOT have `changelogLastRead`:
  - If user has visited site before: treat 5/10/25 as the most-recently read entry.
  - If user is visiting site for the first time: treat all entries as already read.

`changelogLastRead` and `changelogLastReadChangeCount` are updated to match the most recent entry when the dialog is closed. The code implicitly relies on the invariant that the changelog remains sorted in chronological order, and that entries cannot be removed from the changelog, but will not break catastrophically if these invariants are violated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 # incremental build
 tsconfig.tsbuildinfo
 .react-router/
+.vite/
 
 # misc
 .DS_Store
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,6 +8,7 @@
       "name": "xiv-in-the-shell",
       "version": "0.1.0",
       "dependencies": {
+        "@base-ui-components/react": "^1.0.0-alpha.8",
         "@types/jquery": "^3.5.14",
         "@types/react": "^19.1.2",
         "@types/react-dom": "^19.1.3",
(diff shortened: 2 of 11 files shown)


zqsz-xiv:
Requesting your review for localization proofreading @miyehn (as usual, feel free to take your time).

PR #152, opened 2025-05-24
Fixes for crit + DH values

zqsz-xiv:
Yuyuka sent me a bug report via Discord DM that crit buff values were severely over-reporting their effects on auto-crit abilities, like hammers.

For a timeline of 3 hammers under a 10% crit buff (Battle Litany or Chain Stratagem), current main reports their effect as a ~16% buff:
image

The same timeline + stats after this PR reports a more accurate 6%:
image

After scrounging around [xivgear's codebase](https://github.com/xiv-gear-planner/gear-planner/blob/master/packages/xivmath/src/xivmath.ts), I found some easily digestable formulas and found areas where our multipliers for auto-crit/DH buffs were incorrect. I tested these ratios with some short simulations using [our integration with Ama's combat sim](https://colab.research.google.com/github/zqsz-xiv/xivintheshell-ama-sim-notebook/blob/main/in_the_shell_with_ama_sim.ipynb), and encoded the results in a test.

Some results for stacking crit buffs are incorrect, I would guess due to some additive components in the damage formula that we don't account for. Our reported multipliers for these are within ~3% of simulated values, so the difference is not too large.

(FYI @TMTurtle in case you were using these formulas elsewhere)

### DIFF
--- a/src/Controller/Controller.ts
+++ b/src/Controller/Controller.ts
@@ -1461,7 +1461,10 @@ class Controller {
 				buffName = "Radiant Finale";
 			}
 			const getBuffModifiers = () => {
-				if (buff.info.name === BuffType.Dokumori) {
+				if (
+					buff.info.name === BuffType.Dokumori ||
+					buff.info.name === BuffType.ChainStratagem
+				) {
 					return "Debuff Only";
 				} else if (buff.info.name === BuffType.RadiantFinale1) {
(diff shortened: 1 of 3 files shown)


zqsz-xiv:
Incidental fixes:
- bug with combat sim export where dokumori/chain strat weren't applying to boss
- new "levels" field for changelog to avoid using bright color for every change

PR #193, opened 2025-07-10
GameState cleanup

### LINKED ISSUE #186: GameState cleanup
- Allow groups in `autoUpgrade`, `autoDowngrade`, and `replaceIf`
  - These functions explicitly error if trying to replace a skill with itself, but a lot of declarations can be made more ergonomic by allowing self-replacements to be declared (and ignored).
- Create common code for positional and combo potencies
- Create common `WarningType` categories
  - Most jobs emit warnings when a gauge is overcapped, or when a buff is allowed to expire. We should make these generic so the warnings can be specified in a declarative fashion.
- Make sure all code uses appropriate helper functions (`gainStatus`, `tryEnqueueResourceDrop`, `hasResourceAvailable`, `hasResourceExactly`)
- Use common functions for procmode + guaranteed RNG event checks
- Allow undefined values in combineEffects to avoid needing NO_EFFECT
- ~~GCD displays for jobs with inherent haste~~

zqsz-xiv:
Resolves #186.

I found 2 additional bugs in the course of making the listed refactors:
1. I straight up forgot to implement AST's Macro/Microcosmos
2. MNK and NIN did not check the position in a combo tree when applying combo potencies, only that the combo tree itself was active.

I also incidentally made MNK start encounters with Formless Fist active (I meant to do it yesterday but forgot).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
+# secret test corpus
+src/__test__/secret*
+
 # temporary files
 scripts/*_xivapi_cache.csv
 # dependencies
--- a/scripts/generate_job_data.py
+++ b/scripts/generate_job_data.py
@@ -622,7 +622,6 @@ import {{
     MakeAbilityParams,
     MakeGCDParams,
 	MOVEMENT_SKILL_ANIMATION_LOCK,
-	NO_EFFECT,
 	PotencyModifierFn,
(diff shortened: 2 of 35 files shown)


zqsz-xiv:
Also added a secret test corpus script, and detected a GNB combo bug in the old implementation that I think incidentally worked correctly in practice, but might have broken in some weird combo timeout edge cases. My refactors to combo logic made the buggy behavior occur, and I added a proper fix for it.

PR #204, opened 2025-07-23
Replace cast time table with GCD table

zqsz-xiv:
Caro mentioned to me via Discord DM that the FPS tax table tooltip (made for BLM) is confusing on other jobs:
image

Since no job really deals with caster tax now, there's no point in keeping this table around anymore. I've replaced it with a table that adjusts according to sps/sks and inherent haste of the class:
image

image


miyehn:
Rip caster tax even for BLM :0
Code changes look good to me. Although I'm also not experienced in frontend development..
* Should we title the second column "After FPS tax" / "帧率税后" so it doesn't look like there's a few seconds of FPS tax?
* Also should we have a column showing the base GCDs (2.5s, 3s, etc.) before applying sps/sks and any inherent haste? Or are they obvious enough to the users?

PR #212, opened 2025-08-04
Clean up timeline canvas logic

zqsz-xiv:
The timeline canvas is now split into 3 layers, with no global variables. State management is more react-y than before, but still a bit jank due to some issues with callbacks reading stale values.

There are 3 advantages to this approach compared to the current codebase:
1. Easier to reason about click/drag state; previous implementation had some weird behavior due to `testInteraction` handlers always running a frame after state updates were performed.
2. No more global state, so the timeline canvas essentially functions as an independent component. No more  `swapCtx` hack for image export.
3. Way fewer redraws. Performance isn't an issue, but the canvas (and other components affected by controller updates) are redrawn way less often with this PR.

The primary downside is that a lot of state management is done somewhat manually, and different `useEffect` hooks are now responsible for explicitly performing redraws and certain state updates. I personally find this easier to reason about than before because react variables are immutable within an update cycle, but YMMV.


miyehn (commented):
I was about to say the overall changes look good to me but then I tested it locally and the interaction is still buggy, likely because of dpr:
canvas-dpr
So you might be able to reproduce if you test on a different monitor, or move the browser window from one monitor to another :0

I also encountered a case where the canvas (maybe one of the layers) not getting completely cleared and the mouse hover tip just keeps drawing on top of each other. But I forgot the steps to reproduce so don't have a gif.
I wonder if in `clearCtx` we can just call `ctx.clearRect(0, 0, ctx.canvas.width, ctx.canvas.height)` to always clear the entire canvas?

zqsz-xiv:
> I also encountered a case where the canvas (maybe one of the layers) not getting completely cleared and the mouse hover tip just keeps drawing on top of each other. But I forgot the steps to reproduce so don't have a gif.

I got a reproduction by selecting something and then using ctrl-+/ctrl-- to change the zoom:
image

Seems to also be DPR-related as well.

zqsz-xiv:
I thought it would take longer to fix, but it turns out I just needed to move some `scale` calls around to ensure all redraw and clear calls were using the correct DPR. Should be fixed now.

> I wonder if in clearCtx we can just call ctx.clearRect(0, 0, ctx.canvas.width, ctx.canvas.height) to always clear the entire canvas?

I think this won't be sufficient because the left boundary can be negative.

zqsz-xiv:
Hm, there's probably some kind of event listener we need to hook up to for the moving to another monitor case; even though the `dpr` is in the sensitivity list of the redraw effect I'm not sure if it actually triggers in all cases where it needs to since it's just a normal variable rather than something managed by react. I'll test that case out tomorrow before merging.

PR #218, opened 2025-08-10
Misc adjustments

miyehn:
Readme: me and Eshiya we have enough credit in the about section already (and the tool's deviated enough from the original BLM in the Shell), so I'm removing us and putting shanzhe there

Timeline editor:
* Title and footer background color: I changed them to `bgMediumContrast` because it has text on top and `bgHighContrast` might be a bit too close to text color
* Removing alternating row background: I think the table grid already serves the purpose of differentiating all the rows (and each row here is pretty short anyway), so alternating background might be redundant. This is what it looks like after the change:image
If we have concerns, we can make the grid lines more prominent (`bgHighContrast`)
Or keep the alternating background and potentially remove the grid, but might want to tone it down a bit, closer to the contrast level of `colors.timeline.tracks`, so it guides the users but doesn't draw too much attention
(Timeline tracks uses alternating background because each row is reallly long, even grid lines might not be enough for guiding users)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,9 +6,7 @@ https://xivintheshell.com
 
 Let us know if you encounter bugs and/or have questions or concerns.
 
-Initial idea: Galahad Donnadieu @ Exodus (`ukitaeshiya`)  
-Initial developer: Ellyn Waterford @ Sargatanas (`miyehn`)  
-Currently maintained by: [@developers](https://github.com/orgs/xivintheshell/teams/developers) of [FFXIV in the Shell](https://github.com/xivintheshell).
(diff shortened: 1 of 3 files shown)


miyehn:
I guess... yeah it look a little strange if I remove the borders, so how about this:
image
I also added a bit more space between the columns

zqsz-xiv:
Coloring looks better, but the left/right spacing feels awkward, especially since it doesn't match the vertical spacing. My thinking is that they should look like spreadsheet cells, which justify text very close to the boundaries by default.

miyehn:
image
ok I put the original padding back

noloerino (dismissed):

zqsz-xiv (approved):

zqsz-xiv:
ignore me accidentally not hitting the account switcher

